Issue #21, opened 2018-07-25
Missing test suite


ltfschoen:
Propose to add test suite similar to that used on polkadot-js/apps

Issue #9, opened 2018-07-02
Calculate average block time


maciejhirsz:

ltfschoen:
It appears as though the functionality of calculating the average block time already been implemented. Although there is duplication:
* https://github.com/polkadot-js/dotstats/blob/master/packages/backend/src/Node.ts#L180
* https://github.com/polkadot-js/dotstats/blob/master/packages/backend/src/Chain.ts#L112

The average blocktime is updated in the frontend here:
* https://github.com/polkadot-js/dotstats/blob/master/packages/frontend/src/Connection.ts#L123
* https://github.com/polkadot-js/dotstats/blob/master/packages/frontend/src/state.ts#L8
* https://github.com/polkadot-js/dotstats/blob/master/packages/frontend/src/components/Chain.tsx#L86

maciejhirsz:
Ye, the average block time on Node isn't being used right now (it's not displayed anywhere). Note that average block time for node can be different than that of the chain as a whole (which is getting block times from different nodes, whichever is fastest at any given block height).

PR #22, opened 2018-07-25
fix: Relates to #9. Reuse calculation of average block time

ltfschoen:
* **Issues**
  * Unable to import `blockAverage` into Chain.ts or Node.ts due to Issue #21.
  * TypeScript says `[ts] Module '"/Users/Me/code/blockchain/clones/paritytech/dotstats/packages/common/build/index"' has no exported member 'blockAverage'.`

### DIFF
--- a/packages/backend/src/Node.ts
+++ b/packages/backend/src/Node.ts
@@ -1,6 +1,7 @@
 import * as WebSocket from 'ws';
 import * as EventEmitter from 'events';
-import { noop, timestamp, Maybe, Types, idGenerator } from '@dotstats/common';
+
+import { noop, timestamp, Maybe, Types, idGenerator, blockAverage } from '@dotstats/common';
 import { parseMessage, getBestBlock, Message, BestBlock, SystemInterval } from './message';
 import { locate, Location } from './location';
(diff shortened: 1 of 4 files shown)


packages/common/src/block.ts
```diff
@@ -0,0 +1,19 @@
+import { Maybe, Types } from '@dotstats/common';
+
+export function blockAverage(blockTimes: Array<number>): Maybe<Types.Milliseconds> {
```
maciejhirsz: You never return a nullish value, so `Maybe` around return type isn't necessary.

packages/backend/src/Node.ts
```diff
@@ -177,22 +177,8 @@ export default class Node {
...
-      }
-    }
-
-    if (accounted === 0) {
-      return 0;
-    }
-
-    return sum / accounted;
+  public get average(): Types.Milliseconds {
+    return blockAverage(this.blockTimes);
```
maciejhirsz: `blockAverage` returns `Maybe`, typescript should complain here (though that should be fixed on the function itself, comment below).

Issue #16, opened 2018-07-21
Allow defining a list of watched nodes


chevdor:
As the list grow, telemetry is nice to monitor some node of interest.
As a user, I would like to define some 'watched' nodes that stay pinned on top of the list

ltfschoen:
As suggested by @maciejhirsz, propose a similar approach to the click to "pin" functionality shown at https://ethstats.net/

chevdor:
works for me!

PR #27, opened 2018-07-28
WIP: #16 allow defining list of watched nodes

ltfschoen:
- [X] Add an pinning icon column (that the user may click, ultimately to to pin a Polkadot Node at the top)
- [X] Clicking pinned icon toggles parent component state and passes down updated state via props
- [X] Tried to fix issue where Node component UI does not re-render when parent component state changes. Tried converting Row component into a class and adding lifecycle method shouldComponentUpdate to Row to update the component if the nodesPinned prop changes so that the colour of the pinned icon changes from white to red but fails to compile due to error `A maximum of 1 class per file is allowed.` since both `Row` and `Location` are both classes in the file Node.tsx. 
- [X] Try moving Row component it into a separate file since only one class allowed per file.
Fixed using this approach: https://stackoverflow.com/questions/31983209/using-namespace-spread-over-multiple-files-in-typescript
- [X] Store a hash of the node id and a boolean value of whether it's pinned or not in the parent component or ideally in local storage
- [ ] After adding an item as pinned, if you remove it, then after you refresh the page it'll still stay shown as pinned...
- [ ] Sometimes the Node Id's that are passed via props from state change to different numbers so the Node Ids shown in the UI change and don't have a pinned state stored. Require Issue #32 https://github.com/polkadot-js/dotstats/issues/32 to be implemented so we store the public key of the node in the state and local storage instead 
- [ ] Remove console.logs


maciejhirsz:
This will be heavily affected by #34. I can help with merging it back into this branch.

maciejhirsz:
FYI: This PR was pretty stale and fixing conflicts would have been more bothersome than implementing it fresh, especially after I added a `Persistent<T>` abstraction for settings that now handles all the heavy lifting for pins in #48.

PR #111, opened 2019-02-18
Created a Dockerfile & docker-compose so theres no need to run in separate terminals

marbar3778:
Addition of a docker file to feed a docker-compose file to be able to easier serve the application. 

Changes in UI are copied in via volumes so if you are developing then you can still use docker-compose up.

- [x] update Readme


marbar3778:
[clabot:check]

PR #123, opened 2019-03-14
Add network state to backend

gterzian:


packages/backend/src/message.ts
```diff
@@ -53,6 +53,7 @@ export interface SystemConnected {
 
 export interface SystemInterval extends BestBlock {
   msg: 'system.interval';
+  network_state: Types.NetworkState;
```
maciejhirsz: This should be a `Maybe<Types.NetworkState>`. Older versions of Substrate won't be sending those, Backend shouldn't flip on `undefined`.

packages/common/src/types.ts
```diff
@@ -24,6 +24,7 @@ export type MemoryUse = Opaque<number, 'MemoryUse'>;
 export type CPUUse = Opaque<number, 'CPUUse'>;
 export type BytesPerSecond = Opaque<number, 'BytesPerSecond'>;
 export type NetworkId = Opaque<string, 'NetworkId'>;
+export type NetworkState = string;
```
maciejhirsz: Would keep it in line with `Opaque<string, 'NetworkState'>`, so that type inspection tells you what the value is (this is most useful for passing messages to the UI, which is done with plain arrays without field names).

packages/backend/src/Node.ts
```diff
@@ -79,6 +80,7 @@ export default class Node {
     this.address = address;
     this.authority = authority;
     this.networkId = networkId;
+    this.networkState = "Not received yet";
```
maciejhirsz: I think I'd prefer the type to be a `Maybe<Types.NetworkState>` here with default `null` (analog to other optional values, like `location`). If there were to be a default string value, then it should be a string value that can be parsed as JSON.

gterzian:
@maciejhirsz Ok, comments addresssed...

gterzian:
@maciejhirsz Yes on the front-end in a separate PR. And I think we discussed your other question in the chat...

PR #128, opened 2019-03-20
Add network state to UI

gterzian:
@maciejhirsz I'm running into some issues, and I think it's best to discuss them in here...

### DIFF
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -1,7 +1,10 @@
+import * as http from 'http';
+import * as url from 'url';
 import * as WebSocket from 'ws';
 import Node from './Node';
 import Feed from './Feed';
 import Aggregator from './Aggregator';
+import {Types} from '@dotstats/common';
 
 const WS_PORT_TELEMETRY_SERVER = Number(process.env.TELEMETRY_SERVER || 1024);
 const WS_PORT_FEED_SERVER = Number(process.env.FEED_SERVER || 8080);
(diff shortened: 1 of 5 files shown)


packages/backend/src/index.ts
```diff
@@ -47,3 +50,29 @@ logClients();
...
+http.createServer((request, response) => {
+  const incoming_url = request.url || "";
+  const parsed_url = url.parse(incoming_url, true);
+  const path = parsed_url.path || "";
+  if (path.startsWith("/network_state/")) {
+    const [chain_id, str_node_id] = path.split('/').slice(2);
+    console.log(chain_id, str_node_id);
+    const node_id = Number(str_node_id);
+    const chain = aggregator.getChain(chain_id as Types.ChainLabel);
+    const nodeList = Array.from(chain.nodeList());
```
gterzian: This, for some reason, is empty even after the node has connected to telemetry. Any ideas?

packages/frontend/src/components/List/Row.tsx
```diff
@@ -281,6 +281,13 @@ export class Row extends React.Component<Row.Props, Row.State> {
       setting: 'blocklasttime',
       render: ({ blockTimestamp }) => <Ago when={blockTimestamp} />
     },
+    {
+      label: 'NetworkState',
+      icon: networkIdIcon,
+      width: 100,
+      setting: 'networkstate',
+      render: ({ id }) => React.createElement('a', {href: `/network_state/0/${id}/`, target: "_blank"}, "Network state"),
```
gterzian: What would be the best way to get the `chainLabel` here?

maciejhirsz: You could use `getHashData().chain` (from utils), which will pull it from the url. You might want to url encode it (and decode it on the backend).

packages/backend/src/index.ts
```diff
@@ -47,3 +50,29 @@ logClients();
...
+    console.log(nodeList)
+    const node = nodeList.filter((node) => node.id == node_id)[0];
+    console.log(node)
+    let json;
+    if (node) {
+      json = JSON.stringify(node.networkState);
```
maciejhirsz: Isn't `networkState` a JSON string already?

gterzian:
@maciejhirsz Ok, comments addressed!

maciejhirsz:
Cool, will deploy it tomorrow with nginx proxy to the new port.

PR #134, opened 2019-04-03
Display state of Grandpa

cmichi:
Addresses https://github.com/paritytech/substrate/issues/422.

The PR on Substrate's side is https://github.com/paritytech/substrate/pull/2198.

You can view the interactive visualization [here](http://substrate.creal.de:3000/#consensus/Local%20Testnet) (ping me if it is not running). 

Should look roughly like this:
vis7


packages/frontend/public/index.html
```diff
@@ -12,6 +12,7 @@
         color: #111;
       }
     </style>
+    <script type="text/javascript" src="./Jdenticon.min.js"></script>
```
maciejhirsz: Can we find a react component library that we can pull in as a dependency in package.json in the FE?

packages/common/src/types.ts
```diff
@@ -27,7 +27,42 @@ export type NetworkId = Opaque<string, 'NetworkId'>;
...
+
+export declare type Authority = {
+  Address: Address,
+  NodeId: Maybe<NodeId>,
+  Name: Maybe<NodeName>,
+};
+export declare type Authorities = Array<Address>;
+export declare type AuthoritySetId = Opaque<number, 'AuthoritySetId'>;
+export declare type ConsensusInfo = {
+  BlockNumber: ConsensusView;
```
maciejhirsz: I assume this is meant to be mapping. This creates a type definition for an object literal with a field name `BlockNumber`, not a mapping of block numbers to views.

In general, I prefer to use [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) for mappings, it has some advantages over object literals (most importantly, it allows for keys to be any values, it does not cast them to strings in order to hash them). So this would be:

`export type ConsensusInfo = Map<BlockNumber, ConsensusView>;`

That said mappings like these aren't very good candidates for being transmitted over the wire, as they can grow quite large. I'd prefer doing similar thing to what I'm doing with nodes: there are separate states in backend and frontend, and the backend is sending updates made to the state to the frontend without sending the whole thing over.

packages/frontend/public/index.html
cmichi: Unfortunately not, I had searched, but there was no working react library at the time when I started implementing this.

packages/common/src/types.ts
cmichi: > This creates a type definition for an object literal with a field name BlockNumber, not a mapping of block numbers to views.

Ahh, I see.

> […] the backend is sending updates made to the state to the frontend without sending the whole thing over

That's also the way it works in this PR (https://github.com/paritytech/substrate-telemetry/pull/134/files#diff-4e01c1b28b795b06884aafdfd8e1caebR199) ‒ only the blocks for which the consensus info changed since the last broadcast are propagated to the frontend. 

> mappings like these aren't very good candidates for being transmitted over the wire, as they can grow quite large.

This PR keeps only a fixed amount of blocks in the map (https://github.com/paritytech/substrate-telemetry/pull/134/files#diff-4e01c1b28b795b06884aafdfd8e1caebR10), hence it grows only to a fixed size.

> I'm happy to help implementing the changes, but it will have to wait till after Sub0

:+1:

packages/frontend/src/AfgHandling.ts
```diff
@@ -0,0 +1,265 @@
...
+      }
+
+      this.initialiseConsensusView(consensusInfo, height, own, other);
+
+      const index =
+        consensusInfo.findIndex(([blockNumber,]) => blockNumber === height);
+      if (index === -1) {
+        cont = false;
+        break;
+      }
```
maciejhirsz: Shorter way to write the whole loop (the index check in the middle is superficial as noted above):

```ts
for (const [height, consensusView] of consensusInfo) {
  if (height >= to) {
    contine;
  }

  this.initialiseConsensusView(consensusInfo, height, own, other);

  const cont = f ? f(height, consensusView) : true;
  if (!cont) {
    break;
  }
}
```

It might make sense to extract a function that builds up `Types.ConsensusView` that doesn't need to index into `consensusInfo`, that is:

```ts
this.initialiseConsensusView(consensusInfo, height, own, other);
```

Could become:
```ts
this.intialiseConsensusViewByRef(consensusView, own, other);
```

The former can then use the latter internally to keep things DRY.

packages/frontend/src/AfgHandling.ts
```diff
@@ -0,0 +1,265 @@
...
+    const index = consensusInfo.findIndex(([blockNumber,]) => blockNumber === height);
+    const [, consensusView] = consensusInfo[index];
+
+    for (const o in obj) {
+      if (obj.hasOwnProperty(o)) {
+        consensusView[addr][voter][o] = obj[o];
+      }
+    }
+
+    if (index !== -1) {
```
maciejhirsz: `index` is never `-1` at this point.

packages/frontend/src/AfgHandling.ts
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  private updateConsensusInfo(
+    height: Types.BlockNumber,
+    addr: Types.Address,
+    voter: Types.Address,
+    obj: object,
+  ) {
+    const consensusInfo = this.getState().consensusInfo;
+    this.initialiseConsensusView(consensusInfo, height, addr, voter);
```
maciejhirsz: Might be better to return `ConsensusView` instead of `BlockNumber` from `initialiseConsensusView` and avoid having to do the search through the array again.

packages/frontend/src/AfgHandling.ts
```diff
@@ -0,0 +1,265 @@
...
+    const [, consensusView] = consensusInfo[index];
+
+    for (const o in obj) {
+      if (obj.hasOwnProperty(o)) {
+        consensusView[addr][voter][o] = obj[o];
+      }
+    }
+
+    if (index !== -1) {
+      consensusInfo[index] = [height, consensusView];
```
maciejhirsz: This is a no-op.

packages/frontend/src/App.tsx
```diff
@@ -53,6 +53,10 @@ export default class App extends React.Component<{}, State> {
       status: 'offline',
       best: 0 as Types.BlockNumber,
       finalized: 0 as Types.BlockNumber,
+      consensusInfo: new Array(),
```
maciejhirsz: `as Types.ConsensusInfo`

packages/frontend/src/AfgHandling.ts
cmichi: Ah thanks, great catch! But I think you meant the other branch, which is supposed to append the block data `[height, consensuView]` at the beginning if it's not already in the `consensusInfo`. This is never necessary because of `initialiseConsensusView`, which already does that. I removed this branch now.

packages/frontend/src/AfgHandling.ts
cmichi: See answer above.

packages/frontend/src/AfgHandling.ts
cmichi: Good point, I changed it :+1:.

packages/backend/src/Node.ts
```diff
@@ -283,6 +311,57 @@ export default class Node {
...
+  private onAfgAuthoritySet(message: AfgAuthoritySet) {
+    const {
+      authority_set_id: authoritySetId,
+      hash,
+      number,
+    } = message;
+
+    // we manually parse the authorities message, because the array was formatted as a
+    // string by substrate before sending it.
+    const authorities = JSON.parse(String(message.authorities)) as Types.Authorities;
```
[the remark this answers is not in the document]
cmichi: Do I understand you right that we should be able to remove the `serde` serialization happening [here](https://github.com/paritytech/substrate/blob/master/core/finality-grandpa/src/lib.rs#L511-L518)?

(I had tried it naively without looking any further into it and ran into `the trait 'slog::Value' is not implemented for 'std::vec::Vec<std::string::String>'`.)

cmichi:
@maciejhirsz Thanks for the comments, they were all really informative! I updated the PR with your comments addressed.

packages/backend/src/Node.ts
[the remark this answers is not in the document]
maciejhirsz: Yes, I think that was fixed recently.

PR #139, opened 2019-04-19
Make chain header scrollable

### LINKED ISSUE #92: Unable to horizontal scroll to view all Substrate and Polkadot testnet chain tabs
Screenshot showing Github logo overlapping the testnet chain tabs screen shot 2019-02-01 at 8 18 36 am

Mischi:
Hi.

This PR makes the chain header scrollable and will therefore display all chains instead of just a few and fixes #92.

I did the following changes:
- Make the chain header scrollable and therefore display all chains
- Limit the width of each chain header to 150px
   - Display ellipsis if chain label is not completely visible.
   - The full chain label is visible on mouse hover by setting the `title` attribute
- Make the chain header height dynamic to make some room for the horizontal scroll bar if one is needed
- Move the Github ForkMe! to the left side so that it is always visible 

I think merging this PR would improve the situation for some time but I also think that the chain header concept has to be reworked completely once the number of chains grows considerably.  

Thanks,
Fabian

substrate-telemetry-scrollable-header

### DIFF
--- a/packages/frontend/src/components/Chain/Chain.css
+++ b/packages/frontend/src/components/Chain/Chain.css
@@ -1,3 +1,8 @@
+.Chain {
+  display: flex;
+  flex-direction: column;
+}
+
 .Chain-header {
   width: 100%;
   height: 108px;
@@ -17,11 +22,7 @@
 }
 
 .Chain-content-container {
-  position: absolute;
-  left: 0;
-  right: 0;
-  bottom: 0;
-  top: 148px;
+  flex: 1;
 }
 
 .Chain-content {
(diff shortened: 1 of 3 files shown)


maciejhirsz:
Thanks for this @Mischi! I've had a couple ideas on how to solve this, I'm not settled on anything in particular. I think it would make sense to just crop the tabs leaving top, and then have an icon that opens a page with all other chains sending telemetry data. I'll leave this open and merge it in case I (or someone else) builds on top of this, which is likely :).

maciejhirsz:
Closing in favor of #157.

PR #153, opened 2019-06-14
Add CENNZnet logo

xlc:


maciejhirsz:
@xlc thanks, this is now deployed

PR #160, opened 2019-07-01
feat: Don't count stale nodes

maciejhirsz:


packages/backend/src/Chain.ts
```diff
@@ -80,8 +82,13 @@ export default class Chain {
   public removeNode(node: Node) {
     node.events.removeAllListeners();
 
+
```
cmichi: Accidental newline?

PR #168, opened 2019-08-06
Dockerfile: add dependency on nodejs 10

gabreal:
docker build will fail like this:
```
% docker build --tag parity/substrate-telemetry:latest .
Sending build context to Docker daemon  4.423MB
Step 1/7 : FROM node:alpine
 ---> d97a436daee9
Step 2/7 : WORKDIR /app
 ---> Using cache
 ---> 7c3dda0c34df
Step 3/7 : RUN apk update && apk add python g++ make openssh git bash pdftk
 ---> Running in 878af67ffb1f
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/community/x86_64/APKINDEX.tar.gz
v3.9.4-110-gf23de6bf75 [http://dl-cdn.alpinelinux.org/alpine/v3.9/main]
v3.9.4-108-gfec49fe212 [http://dl-cdn.alpinelinux.org/alpine/v3.9/community]
OK: 9774 distinct packages available
  pdftk (missing):
ERROR: unsatisfiable constraints:
    required by: world[pdftk]
The command '/bin/sh -c apk update && apk add python g++ make openssh git bash pdftk' returned a non-zero code: 1
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,16 @@
-FROM node:alpine
+FROM node:10-alpine
 
 WORKDIR /app
 
-RUN apk update && apk add python g++ make openssh git bash pdftk
-RUN export PYTHONPATH=${PYTHONPATH}:/usr/lib/python2.7
+RUN apk add --no-cache python make g++
 
-COPY ./package.json yarn.lock ./
+COPY ./scripts ./scripts
+COPY ./backend ./backend
+COPY ./packages ./packages
+COPY ./package.json ./yarn.lock ./tsconfig.json ./
 
 RUN yarn 
 
-COPY . .
 
 # Frontend is exposing 3000
 # Backend is exposing 8080


Dockerfile
```diff
@@ -1,15 +1,16 @@
...
+FROM node:10-alpine
 
 WORKDIR /app
 
-RUN apk update && apk add python g++ make openssh git bash pdftk
-RUN export PYTHONPATH=${PYTHONPATH}:/usr/lib/python2.7
+RUN apk add --no-cache python make g++
 
-COPY ./package.json yarn.lock ./
+COPY ./scripts ./scripts
```
TriplEight:
```suggestion
COPY ./scripts scripts
```

Dockerfile
```diff
@@ -1,15 +1,16 @@
...
+COPY ./backend ./backend
```
TriplEight:
```suggestion
COPY ./backend backend
```

Dockerfile
```diff
@@ -1,15 +1,16 @@
...
+COPY ./packages ./packages
```
TriplEight:
```suggestion
COPY ./packages packages
```

gabreal:
works for me:


%  docker run -p 3000:3000 parity/substrate-telemetry:latest yarn start:frontend
yarn run v1.17.3
$ scripts/start-frontend.sh
npm WARN lifecycle The node binary used for scripts is /tmp/yarn--1565167647395-0.16855598440807218/node but npm is using /usr/local/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.

> @dotstats/frontend@0.1.0 start /app/packages/frontend
> react-scripts-ts start

Starting type checking and linting service...
Using 1 worker with 2048MB memory limit
Watching: /app/packages/frontend/src
Starting the development server...

ts-loader: Using typescript@2.9.2 and /app/packages/frontend/tsconfig.json
no-unused-variable is deprecated. Since TypeScript 2.9. Please use the built-in compiler checks instead.
Compiled successfully!

You can now view @dotstats/frontend in the browser.

  Local:            http://localhost:3000/
  On Your Network:  http://172.17.0.3:3000/

Note that the development build is not optimized.
To create a production build, use npm run build.


if you bind mount empty or not properly configured node_modules resp packages that might be the reason.

Dockerfile
gabreal: I know it's working the  way it is. If there are reasons to change this I would suggest to change it aligned with the [manual](https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy) which then would copy to absolute paths e.g.
```
COPY ./scripts /scripts/
```

TriplEight: > I know it's working the way it is. If there are reasons to change this I would suggest to change it aligned with the [manual](https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy) which then would copy to absolute paths e.g.
> 
> ```
> COPY ./scripts /scripts/
> ```

I made these suggestins according to the same manual (copying to the WORKDIR)

PR #211, opened 2019-12-09
add Bifrost icon

lurpis:
[clabot:check]


maciejhirsz:
Thanks, this is now deployed

PR #226, opened 2020-02-06
working backend on docker

woss:
refactor of the dockerfiles to be more readable and independent.
full build of the backend (release only) within docker

### DIFF
--- a/backend.Dockerfile
+++ b/backend.Dockerfile
@@ -1,0 +1,11 @@
+FROM rust
+
+WORKDIR /app
+
+COPY ./backend .
+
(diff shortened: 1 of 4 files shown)


woss:
[clabot:check]

maciejhirsz:
Cheers! I've tweaked your Rust a bit. New backend is only using a single port (Node.js had separate ports for submitting telemetry and for subscribing to a feed), so reflecting that as well.

woss:
wow! nice one! yeah, i basically had rust book on one side and fixing that on the other side.

PR #231, opened 2020-02-18
WIP :: Docker images, terraform and dynamic connection

woss:
Changes are:

* no more monorepo since frontend is the only js app
* terraform digital ocean with docker, correct firewall settings , exposing for now only 3 ports
* full docker image builds (yeah docker pull already works)
* docker-compose dev and prod
* connection URL input field for quick and easy connection

PR #238, opened 2020-03-28
Reduce the size of the docker images

chevdor:
This PR shrinks (a lot) the size of the docker images that are produced:
- around 200MB less on the frontend
- more than 1GB less for the backend


chevdor:
I see the solution in #231 goes even further and that's great. I hope we get this other PR split and merged as well.

PR #241, opened 2020-03-30
WIP:: Docker, linter, custom URL for the SUBSTRATE_TELEMETRY_URL, 

woss:
a mistake was made by me and the original PR was closed https://github.com/paritytech/substrate-telemetry/pull/231 this one is for the reference,  @chevdor  and I are working on it and it's easier to track


.travis.yml
```diff
@@ -31,7 +31,5 @@ before_script:
 script: ./test.sh
 
 script:
-  - yarn
-  - yarn build:all
-  - yarn test
-  - cd backend && cargo test
+  - ./scripts/travis-backend.sh
```
chevdor: I think it would be easier to use npm/yarn scripts here.
For instance a new `yarn build:frontend` that, in this case, only goes into the right package and invoke yarn build.

PR #242, opened 2020-03-30
Add prettier and format the code

woss:
* Make prettier works with the tslint 
* Format the code with the new rules
* yarn check and fix scripts
* Travis build and check scripts


packages/frontend/package.json
```diff
@@ -20,21 +20,28 @@
...
-    "blakejs": "^1.1.0"
+    "tslint": "^6.1.0"
   },
   "scripts": {
+    "precommit": "lint-staged",
     "start": "react-scripts-ts start",
     "build": "react-scripts-ts build",
     "test": "echo 'FIXME: FE tests need updating'",
-    "eject": "react-scripts-ts eject"
+    "eject": "react-scripts-ts eject",
```
chevdor: I suggest you add a new `lint` script that does the magic. We can then call the linter with `yarn lint` in Travis before doing anything else.

woss: i wouldn't mess with this pr anymore, there are PRs that are related to this one it would mean changing them too. i can introduce that into the new one

.travis.yml
```diff
@@ -28,10 +28,6 @@ before_script:
   - if [ $TRAVIS_RUST_VERSION = $CLIPPY_TOOLCHAIN ]; then rustup component add clippy-preview --toolchain=$CLIPPY_TOOLCHAIN; fi
   - curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh -s -- -f
 
-script: ./test.sh
 
 script:
   - yarn
-  - yarn build:all
```
chevdor: I don't think it is ok for a PR to stop building the project and remove all tests.

packages/frontend/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "tabWidth": 2,
+  "useTabs": false,
+  "singleQuote": true
+
```
chevdor: The empty lines are not required

packages/frontend/package.json
chevdor: Since this PR is about prettifying and linting, we need a way to start checking whether linting is OK. This yarn script would be the place for that and we can add it to travis too.

packages/frontend/src/AfgHandling.ts
```diff
@@ -43,12 +43,12 @@ export class AfgHandling {
   public receivedFinalized(
     addr: Types.Address,
     finalizedNumber: Types.BlockNumber,
-    finalizedHash: Types.BlockHash,
```
chevdor: Not sure what @maciejhirsz thinks about that one but I find the trailing `,` rather convenient. I would keep them everywhere we can keep them.

chevdor: Ignore that one, my bad.

woss:
@maciejhirsz please review when you can. seems I can't add you as a reviewer

packages/frontend/src/AfgHandling.ts
maciejhirsz: I too like tailing commas, they make git diffs cleaner when adding or removing lines. If there are other things depending on it, I can fix this later so we don't introduce merge conflicts in other open PRs.

.travis.yml
```diff
@@ -28,10 +28,7 @@ before_script:
...
   - curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh -s -- -f
 
-script: ./test.sh
-
 script:
   - yarn
+  - yarn check:all
   - yarn build:all
-  - yarn test
-  - cd backend && cargo test
```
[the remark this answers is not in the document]
woss: added

PR #236, opened 2020-03-23
Add Totem Network logo/icon and update icon selection in frontend

chrisdcosta:
Looking to add these to the public telemetry service.

PR #243, opened 2020-03-30
Restructure the js app

woss:
remove the common code and merge it to the frontend


frontend/src/persist/Persistent.ts
```diff
@@ -13,7 +13,7 @@ export class Persistent<Data> {
     this.key = key;
     this.onChange = onChange;
 
-    const stored = window.localStorage.getItem(key) as Maybe<Stringified<Data>>;
+    const stored = window.localStorage.getItem(key) as any;
```
chevdor, line 16: Why ?

frontend/src/state.ts
```diff
@@ -1,7 +1,7 @@
-import { Types, Maybe, SortedCollection } from '@dotstats/common';
+import { Types, Maybe, SortedCollection } from './common';
 import { Column } from './components/List';
 
-export const PINNED_CHAIN = 'Kusama';
+export const PINNED_CHAIN = 'Kusama CC3';
```
chevdor, line 4: AFAIK, the network was renamed recently from `Kusama CC3` to `Kusama` so this change does not look ok.

frontend/tsconfig.json
```diff
@@ -0,0 +1,38 @@
...
+  "exclude": [
+    "node_modules",
+    "build"
+  ],
+  "include": [
+    "src/**/*.ts",
+    "src/**/*.tsx",
+    "./declarations/**/*.d.ts"
+  ]
+}
```
chevdor, line 38: Missing EOL

frontend/tsconfig.json
```diff
@@ -0,0 +1,38 @@
...
+    "jsx": "react",
+    "rootDir": "src",
+    "moduleResolution": "node",
+    "allowSyntheticDefaultImports": true,
+    "noEmitOnError": false,
+    "pretty": true,
+    "forceConsistentCasingInFileNames": false,
+    "noErrorTruncation": true,
+    "noImplicitReturns": true,
+    "noImplicitAny": false,
```
chevdor, line 22: This used to be `true`. Going back to `false` without an excellent reason would be a regression.

frontend/tsconfig.json
```diff
@@ -0,0 +1,38 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "outDir": "build",
+    "module": "esnext",
```
chevdor, line 5: Used to be `commonjs`. Sure of that change?

frontend/tsconfig.json
```diff
@@ -0,0 +1,38 @@
...
+    "target": "ES2017",
+    "lib": [
+      "es6",
+      "dom"
+    ],
+    "sourceMap": true,
+    "allowJs": true,
```
chevdor, line 12: Do we really need `allowJs` to be `true` now ?

frontend/tsconfig.json
```diff
@@ -0,0 +1,38 @@
...
+    "moduleResolution": "node",
+    "allowSyntheticDefaultImports": true,
+    "noEmitOnError": false,
+    "pretty": true,
+    "forceConsistentCasingInFileNames": false,
+    "noErrorTruncation": true,
+    "noImplicitReturns": true,
+    "noImplicitAny": false,
+    "noImplicitThis": true,
+    "strictNullChecks": false,
```
chevdor, line 24: Used to be true. Why this change?

frontend/tsconfig.json
```diff
@@ -0,0 +1,38 @@
...
+    "suppressImplicitAnyIndexErrors": true,
+    "resolveJsonModule": true,
+    "noUnusedParameters": false
```
chevdor, line 27: From what I see in the code, the frontend was coded the Rust way so unused are marked with `_`.  So instead of allowing unused, we should only allow unused that start with a `_`. I think the old code should pass this way and we prevent regressions in new code.

frontend/src/components/List/Column.tsx
```diff
@@ -58,7 +58,6 @@ import turingIcon from '../../icons/turingnetwork.svg';
 import dothereumIcon from '../../icons/dothereum.svg';
 import katalchainIcon from '../../icons/katalchain.svg';
 import bifrostIcon from '../../icons/bifrost.svg';
-import totemIcon from '../../icons/totem.svg';
```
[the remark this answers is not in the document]
chevdor: This was added a few days ago in:
- #236
- #239

README.md
```diff
@@ -20,35 +22,45 @@ cargo build --release
...
 BIND=0.0.0.0 telemetry
 ```
 
 ### Terminal 2 - Frontend
-```
-yarn start:frontend
+
+```sh
+cd frontend
+yarn
```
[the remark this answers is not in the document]
woss: as the whole thing moved away from the yarn `workspaces`  we need to manually install the deps in the frontend itself.  i don't see why a user cannot just co to the `frontend` and `yarn && yarn start`. But for the sake of the fast start there is a script in the root `start` which starts the frontend.

frontend/package.json
```diff
@@ -25,8 +25,8 @@
     "blakejs": "^1.1.0",
     "husky": "^4.2.3",
     "lint-staged": "^10.1.0",
-    "react": "16.4.0",
-    "react-dom": "16.4.2",
+    "react": "^16.13.1",
```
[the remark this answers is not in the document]
woss: no we cannot, then we need to extract ConnectionInput which is built using the hooks, and hooks are not available in the `16.4.2`

frontend/src/components/ConnectionInput/ConnectionInput.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
+
```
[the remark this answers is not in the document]
woss: checks passing, so yes linter is ok with that. grouping ?  maybe next time or some automatic way of doing it

frontend/tsconfig.json
woss: because of this `substrate-telemetry/frontend/src/AfgHandling.ts
(16,10): 'receivedAuthoritySet', which lacks return-type annotation, implicitly has an 'any' return type.`

woss: and potentially other issues when you solve that. setting that to false is nothing major. if we put that back code will break. also deleting the yarn.lock from frontend will break it even on master. try it.

woss: even before all my big prs

frontend/tsconfig.json
woss: back to false

frontend/tsconfig.json
woss: the moment i opened the project i saw a strongly typed background of a person who did this. the flag was deprecated.

frontend/tsconfig.json
chevdor: "Implicit any" issues are only showing up in ConnectionInput.tsx => specifying the 2 missing types
```
  function handleChange(e: React.ChangeEvent<HTMLInputElement>) {... }
  function handleSubmit(e: SyntheticEvent) { ... }
```

frontend/tsconfig.json
chevdor: It requires only a simple fix in `ConnectionInput.tsx` to be set to true

frontend/tsconfig.json
chevdor: Setting `"noUnusedParameters": true` requires the following changes:
- AfgHandling.ts:L116
- AfgHandling.ts:L139
- AfgHandling.ts:L195
- App.tsx:L194
- Consensus.tsx:360
- Sparkline.tsx:82

frontend/package.json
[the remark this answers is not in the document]
chevdor: 👍

README.md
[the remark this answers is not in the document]
chevdor: 👍

frontend/src/components/ConnectionInput/ConnectionInput.tsx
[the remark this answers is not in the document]
chevdor: Yeah.. that hurts my eyes a bit but I admit this is a detail.
That can be fixed later.

frontend/src/components/List/Column.tsx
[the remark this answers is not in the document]
chevdor: This issue should not be forgotten

frontend/tsconfig.json
chevdor: 👍

frontend/tsconfig.json
woss: back to commonjs

frontend/src/Connection.ts
```diff
@@ -17,12 +10,11 @@ import {
...
 import { AfgHandling } from './AfgHandling';
-import { VIS_AUTHORITIES_LIMIT } from '../../frontend/src/components/Consensus';
+import { VIS_AUTHORITIES_LIMIT } from './components/Consensus';
 import { Column } from './components/List';
+import { ACTIONS } from './common/feed';
 
-const { Actions } = FeedMessage;
-
-const TIMEOUT_BASE = (1000 * 5) as Types.Milliseconds; // 5 seconds
+const TIMEOUT_BASE = (1000 * 5) as Types.Milliseconds; // 3 seconds
```
chevdor, line 17: 5s

frontend/src/components/List/Column.tsx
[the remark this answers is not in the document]
maciejhirsz: Yeah, this will need to be re-added.

frontend/src/components/Map/Location.tsx
```diff
@@ -91,7 +91,7 @@ export class Location extends React.Component<Location.Props, Location.State> {
       city,
     } = this.props.node;
 
-    let validatorRow = null;
+    let validatorRow = <div />;
```
maciejhirsz, line 94: Any reason for this change? This will insert an empty `div` inside a `tbody`, which doesn't quite make sense.

frontend/src/state.ts
maciejhirsz: Yup :point_up:

frontend/src/persist/Persistent.ts
maciejhirsz: I'd definitely keep `Maybe<Stringified<Data>>` here.

frontend/src/components/Map/Location.tsx
chevdor: This comes from activating more null check rules.

PR #245, opened 2020-03-30
Better docker image for the backend

chevdor:
This PR brings the work from @woss in a separate smaller PR.

See also #231 and #241


chevdor:
I had to fix the base image to be able to build. I also reduced the size even more.
We are down to around 4MB for the image!!!

image

The produced image can be tested with:
```
docker run --rm -it chevdor/substrate-telemetry-backend:20200406 --help
```

backend/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+ARG PROFILE=release
+WORKDIR /app
+
+COPY . .
+
+RUN cargo build --${PROFILE} --bins --target x86_64-unknown-linux-musl
+
+
+# MAIN IMAGE FOR PEOPLE TO PULL --- small one#
+FROM fredrikfornwall/scratch-with-certificates
```
kirushik, line 15: I'm sorry, but who is Fredrik Fornwall, and why are we relying on his images when building something our users are going to consume?

Meaning: how are we going to make sure that Fredrik's security practices are sufficient to avoid supply chain incidents, which cryptospace had seen [a lot](https://www.zdnet.com/article/cryptocurrency-startup-hacks-itself-before-hacker-gets-a-chance-to-steal-users-funds/)?

woss: I used his image because it was the only i could find that made statically linked binary. Now, when i have more knowledge in the whole blockchain and crypto i agree we should make our own image. I've also discovered (maybe it's wrong) that all other dynamically linked packages require big base images. if that's not true anymore then feel free to remove that image and replace it with some that you think is safer and accomplishes the same result .

kirushik: What are the exact size constraints in this case, and where are they coming from?
Usually, I would assume consumers to already have base images for something something like `ubuntu:latest` (from downloading other images with the same base), but I can imagine you having a particular usecase when this is not true in mind.

Maybe you should try just [`FROM scratch`](https://docs.docker.com/develop/develop-images/baseimages/), if you care about size and can afford building a fully statically-linked binary here?

chevdor: > What are the exact size constraints in this case ?

Smaller is better! Moreover, using `ubuntu:latest` means we switch to a well known base image (that's good) but also a bigger (less good...) one with potentially way more security issues (way less good).

To be honest I was concern as well when I saw the builder base image at first. I first checked if I could find another builder image but that seems to be the only one or one of the few available.

Ultimately, that would be better to use a home made base image, I agree. But as we speak, there is no such an image I am aware of so using the builder from fredrikfornwall sounds acceptable to me. Once we have better, including a process including `security practices are sufficient to avoid supply chain incidents` (which probably should be defined to ensure compliance), a swap would be rather simple.

Testing if that works with scratch as second stage base image would be great and that would be worth a try in another PR.

kirushik: > one with potentially way more security issues (way less good).

Well, at least `ubuntu:latest` (as well as all other official images on Dockerhub) enjoys an attention of a dedicated security team of Dockerhub, and some other security-related practices like [Vulnerability Scanning](https://docs.docker.com/docker-hub/official_images/#official-image-vulnerability-scanning), signed images and defined [Responsible Disclosure process](https://www.docker.com/legal/security).

I really don't think we should rely on unknown third-parties without any explicit commitment to security, be it for build images, or for release ones.

What you can do is investigate what are the dockerfiles used to build the images you like, and then create your own forks or incorporate some changes in the build images [we already have](https://github.com/paritytech/scripts/tree/master/dockerfiles).
Of course this will come with some responsibility for maintaining the images from now on — but, unlike with the unknown third party, we at least have some control over this maintenance.

If the size is indeed that critical for you (and for some reason don't expect the cross-image base layer sharing to help in your scenario), you definitely should start `FROM scratch`, and not use some third-party base layer — for musl-built Rust code making it run without any environment should be pretty easy; you would have to replace the `openssl` usages with `rustls` ones at worst.

chevdor: Yes, I agree and I think this will be great this way but we may not want to block this PR due to this endeavour.

kirushik: We definitely **do want** to block it — this is an official Parity repository, and we shouldn't have "everything goes" approach to security here.

kirushik: You can use your own account for now, if there's a significant time pressure on you in this regard — but please make it explicit that the way you do it is not an official Parity-supported way (and that our general policies like BugBounty are not applied).

chevdor:
@kirushik the issue here is that there is no official or Parity-maintained image that fits the requirements.

kirushik:
Sure, then we need to take care of creating and maintaining one, and taking the images you're building on top of right now as a starting point would be a good idea (and hopefully gets you like 80% there).
I think our DevOps team will be able to help you with figuring out if this should be added to one of our existing images or to a new one.

chevdor:
Cool, let me know once such an image is available.

kirushik:
@chevdor created a https://github.com/paritytech/devops/issues/498 to track the progress; it seems that this PR is blocked on that issue.

chevdor:
That's seems to be a private repo, or a wrong link. I get a 404.

chevdor:
@kirushik regarding the scratch as base image, I could not test earlier (stuck in a big rebase). I will give it a go and see if we can leave without the certificates. If so, that solves half the issue.

chevdor:
Good news, using scratch as base image for the second stage seems to be find. It reduces the size even further (we start talking in kb here...). I pushed the change to this PR.

chevdor:
@kirushik is there an ETA for the new builder you want to make?

kirushik:
> @kirushik is there an ETA for the new builder you want to make?

@General-Beck are you still on the task of adding musl to our build images? What's the ETA?

kirushik:
> @kirushik is there an ETA for the new builder you want to make?

@General-Beck are you still on the task of adding musl to our build images? What's the ETA?

General-Beck:
@kirushik The first test image I think will be ready next week.

chevdor:
@General-Beck provided an image. It is on me now :)

chevdor:
Using the new parity in house image works all fine. The change has been included.

README.md
```diff
@@ -95,9 +95,9 @@ If you don't you can download for you OS here [Docker Desktop](https://www.docke
 docker-compose up --build -d
 ```
 
-- `-d` stands for detach, if you would like to see logs I recommend using [Kitmatic](https://kitematic.com/) or don't use the -d
- - `--build` will build the images and rebuild, but this is not required every time
- - If you want to makes UI changes, there is no need to rebuild the image as the files are being copied in via volumes.
+- `-d` stands for detach, if you would like to see logs i recommend using [Kitmatic](https://kitematic.com/) or don't use the `-d`
```
maciejhirsz, line 98:
```suggestion
- `-d` stands for detach, if you would like to see logs I recommend using [Kitmatic](https://kitematic.com/) or don't use the `-d`
```

PR #246, opened 2020-03-31
Listen by default to the local interface and allows changing it

chevdor:
This PR introduces the following:
- listens to `127.0.0.1:8000` by default instead of the former more open `0.0.0.0:8000`
- allows changing the interface we listen to: `BIND=0.0.0.0` brings back the previous behavior
- document the supported ENV in the code and in the readme


chevdor:
See also #247

PR #247, opened 2020-03-31
Add clap support

chevdor:
This PR adds basic clap support.
It *removes* the use of ENV variables that I introduced in #246 => thus the version bump.

Here are a few sample runs:

## Help:
```
$ ./telemetry --help
Substrate Telemetry Backend 0.2.0
Parity Technologies Ltd. <admin@parity.io>
This is the Telemetry Backend that injects and provide the data sent by Substrate/Polkadot nodes

USAGE:
    telemetry [OPTIONS]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -l, --listen <socket>    This is the socket address Telemetry is listening to. This is restricted localhost
                             (127.0.0.1) by default and should be fine for most use cases. If you are using Telemetry in
                             a container, you likely want to set this to '0.0.0.0:8000' [default: 127.0.0.1:8000]
```

## Version:
```
$ ./telemetry --version
Substrate Telemetry Backend 0.2.0
```

## Default run:
```
$ ./telemetry 
2020-03-31 16:17:15,161 INFO  [actix_server::builder] Starting 4 workers
2020-03-31 16:17:15,163 INFO  [actix_server::builder] Starting server on 127.0.0.1:8000
...
```

## Invalid IP or Port:
```
$ cargo run -- --listen 256.0.0.0:8001
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/telemetry --listen '256.0.0.0:8001'`
error: Invalid value for '--listen <socket>': invalid IP address syntax
```

## Invalid network interface:
```
$ cargo run -- --listen 1.0.0.0:8001
    Finished dev [unoptimized + debuginfo] target(s) in 0.33s
     Running `target/debug/telemetry --listen '1.0.0.0:8001'`
Error: Os { code: 49, kind: AddrNotAvailable, message: "Can\'t assign requested address" }
```

## Run with opts:
```
$ ./telemetry -b 0.0.0.0 -p 8765
2020-03-31 16:17:40,018 INFO  [actix_server::builder] Starting 4 workers
2020-03-31 16:17:40,019 INFO  [actix_server::builder] Starting server on 0.0.0.0:8765
...
```

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -4,22 +4,48 @@ extern crate log;
 use std::net::Ipv4Addr;
 
 use actix::prelude::*;
-use actix_web::{web, App, HttpRequest, HttpResponse, HttpServer, Error};
-use actix_web_actors::ws;
 use actix_http::ws::Codec;
+use actix_web::{web, App, Error, HttpRequest, HttpResponse, HttpServer};
+use actix_web_actors::ws;
+use clap::Clap;
 
-mod types;
 mod aggregator;
 mod chain;
-mod node;
 mod feed;
+mod node;
+mod types;
 mod util;
 
-use node::connector::NodeConnector;
+use aggregator::{Aggregator, GetHealth, GetNetworkState};
 use feed::connector::FeedConnector;
-use aggregator::{Aggregator, GetNetworkState, GetHealth};
-use util::{Locator, LocatorFactory};
+use node::connector::NodeConnector;
 use types::NodeId;
+use util::{Locator, LocatorFactory};
+
+const VERSION: &'static str = env!("CARGO_PKG_VERSION");
+const AUTHORS: &'static str = env!("CARGO_PKG_AUTHORS");
+const NAME: &'static str = "Substrate Telemetry Backend";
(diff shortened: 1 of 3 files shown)


backend/src/main.rs
```diff
@@ -4,22 +4,48 @@ extern crate log;
...
+    )]
+    port: u16,
+
+    #[clap(
+        short = "b",
+        long = "bind",
+        default_value = "127.0.0.1",
+        help = "This is the address of the network interface Telemetry is listening to. This is restricted localhost by default and should be fine for most use cases. If you are using Telemetry in a container, you likely want to set this to '0.0.0.0'"
+    )]
+    bind: String,
```
maciejhirsz: Could try to make this a [`std::net::IpAddr`](https://doc.rust-lang.org/std/net/enum.IpAddr.html), so that if there is any user error in the CLI, the parsing fails at clap level, instead of when trying to spin the server.

chevdor: 💯Let me improve that

chevdor: Fixed using SocketAddr

maciejhirsz: `SocketAddr` includes port, so that would make port redundant.

PR #248, opened 2020-04-01
Fix unit tests

chevdor:
This PR does:
- remove jest
- restore tape tests
- convert tests to TS
- adds test coverage support using istanbul


chevdor:
@maciejhirsz any issue with this PR?

frontend/src/Connection.ts
```diff
@@ -63,8 +63,11 @@ export class Connection {
         clean();
         resolve(null);
       }
-
-      const socket = new WebSocket(Connection.address);
+      const address = localStorage.getItem('connectionURI');
```
maciejhirsz, line 66: Where is this local storage value set? Is this for only for mock connetion injection? This should almost definitely be moved to line 30 along other checks (or the whole thing should be moved to the constructor if you don't want to nest ternarny expressions).

frontend/src/Connection.ts
```diff
@@ -63,8 +63,11 @@ export class Connection {
...
+      if (!address) {
+        resolve(null);
+      }
+      const socket = new WebSocket(address || 'ws://localhost:8000');
```
maciejhirsz, line 70: `Connection.address`

chevdor:
The connection URL stuff are a bit out of scope for the PR but I will make the changes.

frontend/src/Connection.ts
```diff
@@ -48,28 +62,31 @@ export class Connection {
...
-      }
-
-      function onSuccess() {
-        clean();
-        resolve(socket);
-      }
-
-      function onFailure() {
-        clean();
+      if (!Connection.address) {
```
maciejhirsz, line 65: This should never be null, so the check is not necessary.

frontend/src/Connection.ts
```diff
@@ -26,11 +32,19 @@ export class Connection {
...
 
-  private static readonly address =
-    window.location.protocol === 'https:'
-      ? `wss://${window.location.hostname}/feed/`
-      : `ws://127.0.0.1:8000/feed`;
+  private static getAddress(): string {
+    const ENV_URL = 'SUBSTRATE_TELEMETRY_URL';
+
+    if (process.env && process.env[ENV_URL]) return process.env[ENV_URL] as string;
+    if (window.process_env && window.process_env[ENV_URL])
```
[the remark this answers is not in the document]
chevdor: There is no such rule currently and the linter pass. I did a quick check and I don't think tslint even support such a check. I will add the blocks but the linter will no enforce it.

frontend/src/Connection.ts
chevdor: The check actually catch the case of empty strings as well but I will remove it.

PR #256, opened 2020-04-18
Remove Unused Testnet

AurevoirXavier:


chevdor:
@AurevoirXavier @maciejhirsz the fact that CI (Github actions) fail here has nothing to do with the content of this PR. I am removing the labeler (https://github.com/actions/labeler/issues/12) in another PR. I suggest to ignore the failure for this PR.

PR #269, opened 2020-07-13
add docker push on release

lovelaced:
This will add a github action which builds the dockerfiles in CI and pushes them to the paritytech dockerhub.


docker-compose.yml
```diff
@@ -3,13 +3,13 @@ version: "3"
...
-      dockerfile: frontend.Dockerfile
-      context: ./
+      dockerfile: Dockerfile
+      context: ./frontend/
     # Copy in changes to the ui, so no need to rebuild the images.
     volumes:
       - /app/node_modules
       - ./packages:/app/packages
-    command: yarn start:frontend
+    # command: yarn start:frontend
```
s3krit, line 12: Is this supposed to be commented out? Noticed no CMD directive in the Dockerfile

.github/workflows/backend.yml
```diff
@@ -22,3 +22,15 @@ jobs:
...
+      uses: docker/build-push-action@v1 # https://github.com/docker/build-push-action
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+        dockerfile: Dockerfile
+        repository: paritytech/substrate-telemetry-backend-${{ github.event.client_payload.id }}
+        tags: latest
+        add_git_labels: true
+        tag_with_ref: true
+        tag_with_sha: true
```
s3krit, line 36: Check on [this](https://github.com/docker/build-push-action#tag_with_ref) page if you actually need to specify the `latest` tag - from looking at it, it seems to say that when `master` is pushed (and `tag_with_ref` is set to true), the image will automatically be tagged 'latest'.

lovelaced: I want to make sure that the tag this pushes to actually never differs, because I'll be watching a single tag for updates. Maybe i should remove tag_with_ref and tag_with_sha, then. the name of the tag is not set in stone.

docker-compose.yml
lovelaced: This is from @maciejhirsz 's docker-enhancing commit, he might want to do more cleanup that I can pull in. ( @maciejhirsz I was able to successfully build/run both images locally).

docker-compose.yml
```diff
@@ -3,13 +3,13 @@ version: "3"
...
     ports:
       - 3000:3000
```
lovelaced, line 14: this needs to be 3000:80 as per the nginx config running inside the container. we /could/ forgo nginx here, probably?

docker-compose.yml
maciejhirsz: It should be removed. Frontend is served from static files via nginx, `yarn start:frontend` runs dev environment.

docker-compose.yml
```diff
@@ -3,15 +3,14 @@ version: "3"
...
     volumes:
       - /app/node_modules
       - ./packages:/app/packages
-    command: yarn start:frontend
     ports:
-      - 3000:3000
+      - 3000:80
   telemetry-backend:
     build:
       dockerfile: backend.Dockerfile
```
TriplEight, line 16: AFAIU [this](https://github.com/paritytech/substrate-telemetry/blob/master/backend/Dockerfile) is the newer and better image for the backend and the `backend.Dockerfile` shall be removed. 
/cc @chevdor

chevdor: 👍  that's correct.

chevdor: As far as:

>  since we are getting to publish these images in our registry, I presume, it makes sense to update LABEL in Dockerfiles and the main README.md

I am fine with that as long as credits are fair kept. Guys like @woss and myself spent quite some time working on those and that is independant from the new location for hosting the new images.

If you/parity want to take over the maintenance, the best would be syncing with @woss and make the maintainer label a list including him. You could make @woss and myself part of the authors and take over the maintenance label. 

That sounds fair to keep the original authors than to wipe the names of people who initially put things together for the sake of 'cleaning up'.

TriplEight: Of course, I never meant to remove not you, nor @woss from credits. The list is a good idea!
And as for the registry, I think it's more sustainable to set up a CI and push it to a registry on every update automatically rather than manually.

lovelaced: _(for clarity, this will only push to docker hub on tagged commits.)_

lovelaced:
> @lovelaced
> 
> > (for clarity, this will only push to docker hub on tagged commits.)
> 
> Shall we maybe push it on every master change?

No, because we want to update on releases only.

lovelaced:
@TriplEight how should maintenance be handled for the dockerfile(s)? is everything OK as it is now? if so this is ready for merge.

TriplEight:
@lovelaced Yeah, I think it's good to merge and push the first images to the registry. Then we'll take care of fixing a README and metadata, I've struggled to run them in the current state.

lovelaced:
Great, thanks. @maciejhirsz merge at will :)

PR #291, opened 2020-09-29
Upgrade backend to actix-web 3.0

maciejhirsz:


backend/src/node.rs
```diff
@@ -226,7 +226,7 @@ impl Node {
         if let Ok(stringified) = serde_json::from_str::<String>(json) {
             Some(stringified.into())
         } else {
-            Some(json.into())
+            Some(json.to_owned().into())
```
dvdplm, line 229: Some docs would be good on this method, explaining the old vs new node behaviour? How long is it reasonable to support old nodes?

dvdplm (commented):
Looks sane to me, but I have never used Actix so best leave the proper review to the experts.

backend/src/main.rs
```diff
@@ -77,44 +75,59 @@ fn feed_route(
...
+        }
+    }
 }
 
-fn health(
-    aggregator: web::Data<Addr<Aggregator>>,
-) -> impl Future<Item = HttpResponse, Error = Error> {
-    aggregator.send(GetHealth).from_err().and_then(|count| {
-        let body = format!("Connected chains: {}", count);
+/// Entry check for health check monitoring bots
```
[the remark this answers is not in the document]
maciejhirsz: Typo indeed!

backend/src/node.rs
maciejhirsz: New version of `bytes` crate does not have conversion from non-static slices to `Bytes`, so need to clone the JSON dump here. This only happens on the manual network state requests, which happen rarely.

mattrutherford (approved):
LGTM - actix-web is at 3.1.0 which includes some fixes for `NormalizePath` - you might be able to use this to simplify the routing. https://docs.rs/actix-web/3.0.2/actix_web/middleware/normalize/struct.NormalizePath.html

maciejhirsz:
@mattrutherford thanks, I've updated the routes using the route macros now.

dvdplm (approved):

PR #295, opened 2020-10-30
Parachain multiplexing

maciejhirsz:
Companion PR to https://github.com/paritytech/substrate/pull/7463

Expects `id` field on messages (defaults to `0` if absent), messages sent with different `id`s are treated as separate connections and can push the node to different chains.

+ [ ] Check if we need to limit number of chains per node connector.
+ [ ] Check if we need a way to broadcast a single message to all chains per node connector.

Unrelated: fixed continuation frames that sometimes errors on current build.


backend/src/node/connector.rs
```diff
@@ -71,7 +81,9 @@ impl NodeConnector {
     }
 
     fn handle_message(&mut self, msg: NodeMessage, data: Bytes, ctx: &mut <Self as Actor>::Context) {
-        if let Some(chain) = self.chain.as_ref() {
+        let conn_id = msg.id.unwrap_or(0);
```
dvdplm, line 84: I take it `0` lets us be backwards compatible?

backend/src/node/connector.rs
```diff
@@ -22,14 +28,17 @@ pub struct NodeConnector {
...
     /// Aggregator actor address
     aggregator: Addr<Aggregator>,
-    /// Chain actor address
-    chain: Option<Addr<Chain>>,
+    /// Mapping message connection id to addresses of chains for multiplexing
+    /// a node running multiple parachains
+    chains: BTreeMap<ConnId, Addr<Chain>>,
     /// Backlog of messages to be sent once we get a recipient handle to the chain
-    backlog: Vec<NodeMessage>,
+    backlogs: BTreeMap<ConnId, Vec<NodeMessage>>,
```
dvdplm, line 35: Unrelated to this PR, but is it ok that the backlog is unbounded? Maybe we could re-work this into a bounded queue?
EDIT: I see it is bounded to `10` below. Let's document that.

dvdplm (commented):
Looks like this will be fairly painless. :)

backend/src/node/connector.rs
maciejhirsz: Yup yup

backend/src/node/connector.rs
maciejhirsz: Yeah, it's bounded. I'm more worried about the number of `ConnId` being unbounded atm.

backend/src/node/connector.rs
```diff
@@ -98,37 +98,35 @@ impl NodeConnector {
...
+                        _ => (),
+                    }
+
+                    self.aggregator.do_send(AddNode { rec, conn_id, node });
+                } else {
+                    if backlog.len() >= 10 {
+                        backlog.remove(0);
+                    }
+
+                    backlog.push(msg);
```
dvdplm, line 127: So this is the fix here?

maciejhirsz: The issue was that I had a separate backlog and `Addr<Chain>` per `ConnId`, but was using a single `NodeId`, now the `NodeId` is tied to `ConnId`, everything else (using the enum for `ConnMultiplex`) is a refactor to make the intent cleaner and to avoid having 3 separate `BTreeMap`s for no good reason (one for chain address, one for backlog, one for node id).

dvdplm (approved):

insipx (approved):

PR #296, opened 2020-10-31
Prevent nodes out of viewport triggering render

maciejhirsz:
Whenever the backend sends a new block update for a node, the position of that node in the sorted list (using the `SortedCollection` abstraction) needs to be updated. Currently, any alternation of the order prompts the render pipeline in React. The actual table rendered in the UI has already been truncated only to the slice of the list that's visible in the viewport so updates to nodes outside of the viewport produce no changes to the DOM at all, however it would still trigger the VDOM diffing of the entire UI. This was fine while the number of nodes was heavily limited, however we now have multiple chains with multiples of hundreds of nodes on them, the cascade of updates that follows a new block creation can clog the UI, even for seconds.

This PR effectively prevents the React rendering pipeline from being triggered by list ordering changes happening outside of the viewport, there should be no discernable difference in how the UI behaves, aside from being more responsive.

Chromium on my machine is particularly bad, here is a [clip of current live, vs static build of this branch running on live backend](https://maciej.codes/kosz/performance.mp4).

### DIFF
--- a/frontend/src/common/SortedCollection.ts
+++ b/frontend/src/common/SortedCollection.ts
@@ -123,7 +123,7 @@ export class SortedCollection<Item extends { id: number }> {
     this.compare = compare;
   }
 
-  // Set a new focus within which order changes bump `changeRef`
+  // Set a new `Focus`. Any changes to the order of items within the `Focus` will trigger an increase to `changeRef`, which in turn triggers a re-render.
   public setFocus(start: number, end: number) {
     this.focus = { start, end };
   }
--- a/frontend/src/common/SortedCollection.ts
+++ b/frontend/src/common/SortedCollection.ts
@@ -246,7 +246,7 @@ export class SortedCollection<Item extends { id: number }> {
   }
 
   // Get the reference to current ordering state of focused items.
-  public ref(): SortedCollection.StateRef {
+  public get ref(): SortedCollection.StateRef {
     return this.changeRef as SortedCollection.StateRef;
   }
 
--- a/frontend/src/common/SortedCollection.ts
+++ b/frontend/src/common/SortedCollection.ts
@@ -97,14 +97,35 @@ export namespace SortedCollection {
   export type StateRef = Opaque<number, 'SortedCollection.StateRef'>;
 }
 
+interface Focus {
+  start: number;
+  end: number;
+}
+
 export class SortedCollection<Item extends { id: number }> {
+  // Comparator function used to sort the collection
   private compare: Compare<Item>;
+  // Mapping item `id` to the `Item`, this uses array as a structure with
+  // the assumption that `id`s provided are increments from `0`, and that
+  // vacant `id`s will be re-used in the future.
   private map = Array<Maybe<Item>>();
+  // Actual sorted list of `Item`s.
   private list = Array<Item>();
+  // Internal tracker for changes, this number increments whenever the
+  // order of the **focused** elements in the collection changes
   private changeRef = 0;
+  // Marks the range of indicies that are focused for tracking.
+  // **Note:** `start` is inclusive, while `end` is exclusive (much like
+  // `Array.slice()`).
+  private focus: Focus = { start: 0, end: 0 };
 
   constructor(compare: Compare<Item>) {
     this.compare = compare;
+  }
+
+  // Set a new focus within which order changes bump `changeRef`
+  public setFocus(start: number, end: number) {
+    this.focus = { start, end };
   }
 
   public setComparator(compare: Compare<Item>) {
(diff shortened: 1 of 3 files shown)


frontend/src/common/SortedCollection.ts
```diff
@@ -97,27 +97,44 @@ export namespace SortedCollection {
...
   private list = Array<Item>();
+  // Internal tracker for changes, this number increments whenever the
+  // order of the **focused** elements in the collection changes
   private changeRef = 0;
+  // Marks the range of indicies that are focused for tracking.
+  // **Note:** `start` is inclusive, while `end` is exclusive (much like
+  // `Array.slice()`).
+  private focus: Focus = { start: 0, end: 0 };
 
   constructor(compare: Compare<Item>) {
```
emostov, line 122: Will give this a closer read tomorrow, but just a small syntax sugar thing that I like in typescript is that you can drop the body of the constructor here and the member declaration by just having:

```javascript
constructor(private compare: Compare<Item>) {}
```
Read more here: https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties

That being said, having the comment above member declaration is nice and I think this is really just an opinion thing

frontend/src/common/SortedCollection.ts
```diff
@@ -97,27 +97,44 @@ export namespace SortedCollection {
...
     this.compare = compare;
   }
 
+  // Set a new focus within which order changes bump `changeRef`
```
dvdplm, line 126:
```suggestion
  // Set a new `Focus`. Any changes to the order of items within the `Focus` will trigger an increase to `changeRef`, which in turn triggers a re-render.
```

…maybe?

dvdplm (approved):

frontend/src/common/SortedCollection.ts
maciejhirsz: Updated the comment a bit (it doesn't technically trigger anything, plus the abstraction of this particular class knows nothing about the rendering pipeline), and moved it next to `ref()` and `hasChangedSince()` so that intent is clearer.

frontend/src/common/SortedCollection.ts
```diff
@@ -216,6 +239,18 @@ export class SortedCollection<Item extends { id: number }> {
...
   }
 
+  // Set a new `Focus`. Any changes to the order of items within the `Focus`
+  // will increment `changeRef`.
+  public setFocus(start: number, end: number) {
+    this.focus = { start, end };
+  }
+
+  // Get the reference to current ordering state of focused items.
+  public ref(): SortedCollection.StateRef {
```
emostov, line 249: I think it might be more idiomatic to use getters and setters here:

so `public setFocus(start: number, end: number)` would become `set focus(start: number, end: number)`
and `public ref(): SortedCollection.StateRef` would become `get ref(): SortedCollection.StateRef`

Then way either of these accessors where called you would omit the parentheses so it would not look like a function is being invoked.

Unclear to me if there is any real benefit to this other than just syntax preference.

Reference: https://www.typescriptlang.org/docs/handbook/classes.html#accessors

Using accessors requires ECMA 5 (ES2015)  or higher, which I think should be fine since we target ES2017 (I am not to familiar with how this all fits into browser compatibility but I assume babel takes care everything.)

frontend/src/common/SortedCollection.ts
```diff
@@ -131,9 +143,11 @@ export class SortedCollection<Item extends { id: number }> {
 
     this.map[item.id] = item;
 
-    sortedInsert(item, this.list, this.compare);
+    const index = sortedInsert(item, this.list, this.compare);
 
-    this.changeRef += 1;
+    if (index < this.focus.end) {
```
emostov, line 148: Maybe I am missing something but why not `if (index < this.focus.end && index >= this.focus.start)`?

frontend/src/common/SortedCollection.ts
```diff
@@ -147,7 +161,9 @@ export class SortedCollection<Item extends { id: number }> {
     this.list.splice(index, 1);
     this.map[id] = null;
 
-    this.changeRef += 1;
+    if (index < this.focus.end) {
```
emostov, line 164: Same question as my one for the equivalent check in `add`

emostov (approved):
Overall looks good, just had an optional syntax nitpick and a question

frontend/src/common/SortedCollection.ts
maciejhirsz: Removing or adding a node above the focus will shift the entire focus window up or down. It's only the changes below the viewport that have no effect.

frontend/src/common/SortedCollection.ts
maciejhirsz: Ye, `ref` might make sense as a getter, will switch that. I'm not a fan of having a setter that has no corresponding getter (there is no need to check what the focus window currently is, so following YAGNI there is no getter).

PR #300, opened 2020-11-27
Use startup_time from system.connected

maciejhirsz:
Bonus: removed `level` from messages, so they can be dropped in substrate.

### DIFF
--- a/backend/src/node.rs
+++ b/backend/src/node.rs
@@ -33,16 +33,20 @@ pub struct Node {
     location: Option<Arc<NodeLocation>>,
     /// Flag marking if the node is stale (not syncing or producing blocks)
     stale: bool,
-    /// Connected at timestamp
-    connected_at: Timestamp,
+    /// Unix timestamp for when node started up (falls back to connection time)
+    startup_time: Timestamp,
     /// Network state
     network_state: Option<Bytes>,
 }
 
 impl Node {
-    pub fn new(details: NodeDetails) -> Self {
-        Node {
+    pub fn new(mut details: NodeDetails) -> Self {
+        let startup_time = details.startup_time
+            .take()
+            .and_then(|time| time.parse().ok())
+            .unwrap_or_else(|| now());
 
+        Node {
             details,
             stats: NodeStats::default(),
             io: NodeIO::default(),
@@ -52,7 +56,7 @@ impl Node {
             hardware: NodeHardware::default(),
             location: None,
             stale: false,
-            connected_at: now(),
+            startup_time,
             network_state: None,
         }
     }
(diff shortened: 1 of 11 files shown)


backend/src/node/connector.rs
```diff
@@ -161,7 +161,7 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for NodeConnector {
 
         match serde_json::from_slice(&data) {
             Ok(msg) => {
-                // info!("New node message: {}", std::str::from_utf8(&data).unwrap_or_else(|_| "INVALID UTF8"));
+                // log::info!("New node message: {}", std::str::from_utf8(&data).unwrap_or_else(|_| "INVALID UTF8"));
```
niklasad1, line 164: can we remove the comment?

niklasad1 (commented):
LGTM but I don't know all the details :)

dvdplm:
> LGTM but I don't know all the details :)

Yeah, description is a bit lacking here. The context is https://github.com/paritytech/substrate/pull/7463 and https://github.com/paritytech/substrate/pull/7463#discussion_r515011072 I think.

backend/src/node.rs
```diff
@@ -33,16 +33,20 @@ pub struct Node {
     location: Option<Arc<NodeLocation>>,
     /// Flag marking if the node is stale (not syncing or producing blocks)
     stale: bool,
-    /// Connected at timestamp
-    connected_at: Timestamp,
+    /// Unix timestamp for when node started up (falls back to connection time)
```
dvdplm, line 36: Should we really be falling back to connection time here? They are different things. Would switching to an `Option<Timestamp>` be very messy?
If we're ever interested in the time the node connected – and I think we're not – I think it should be a separate field.

dvdplm (changes requested):
I think it would be better to not fall back to `connected_at`

backend/src/node.rs
maciejhirsz: It's a fallback for older version which didn't send the field, which is also what the current functionality is. I believe at this point most nodes should be sending this over, to turn it into an `Option<Timestamp>` requires bumping protocol version at connection handshake with FE, and some changes there to visually account for `null`s.

dvdplm: I see. Can we fall back to `UNIX_EPOCH` or some other value "zero-ish" value?

maciejhirsz: This shows up as Uptime in the FE, I don't think showing 50 years of uptime is better than showing the time the node connected to telemetry at as fallback :)

dvdplm (approved):
Lovely, thank you! :)

niklasad1 (approved):

PR #306, opened 2021-01-07
Wrap message's content into node's message

cecton:
Wraps the node's message content into another object so the ID of the connection is separated from the actual payload.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,8 +1,12 @@
 root = true
-
 [*]
 charset = utf-8
-indent_style = space
-indent_size = 2
 trim_trailing_whitespace = true
 insert_final_newline = true
+indent_style = space
+
+[frontend/**]
+indent_size = 2
+
+[backend/**]
+indent_size = 4
--- a/backend/src/chain.rs
+++ b/backend/src/chain.rs
@@ -319,12 +319,12 @@ impl Handler<UpdateNode> for Chain {
     fn handle(&mut self, msg: UpdateNode, _: &mut Self::Context) {
         let UpdateNode { nid, msg, raw } = msg;
 
-        if let Some(block) = msg.details.best_block() {
+        if let Some(block) = msg.payload.details.best_block() {
             self.handle_block(block, nid);
         }
 
         if let Some(node) = self.nodes.get_mut(nid) {
-            match msg.details {
+            match msg.payload.details {
                 Details::SystemInterval(ref interval) => {
                     if interval.network_state.is_some() {
                         if let Some(raw) = raw {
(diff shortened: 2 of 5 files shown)


maciejhirsz:
There should be a way to make it backwards compatible with a [`#[serde(untagged)]`](https://serde.rs/container-attrs.html#untagged) enum, I can look into it next week.

dvdplm:
A PR description would be good.

FWIW I don't mind this as is, I like the explicitness of `V1`/`V2`.

cecton:
(I keep it in draft because I'm still testing)

cecton:
Ok I made a change to move the timestamp (ts) to the message.

I think this is better, you can clearly see the code in substrate-telemetry is easier. But there is something weird: `ts` wasn't use in the first place!! 😆  Should I remove it or do you plan to do something with it at some point?

I think this is better ALSO because for the code in substrate it removes some complexity for injecting the `ts` in the messages. But since it's not used at all... maybe we should consider removing it entirely?

Last but not least: if we do keep this new implementation I suggest I refactor the name of the struct Details to Payload to match the actual tag. But it's fine and it works as it is now.

Waiting for your feedback on this

maciejhirsz:
> Should I remove it or do you plan to do something with it at some point?

I think it's fine to remove the timestamp, it will shave a bit of time parsing too. As long as it's being sent by the nodes, we can add it back in the future should we need it.

> Last but not least: if we do keep this new implementation I suggest I refactor the name of the struct Details to Payload to match the actual tag. But it's fine and it works as it is now.

Renaming sounds like a good idea to me.

dvdplm:
> As long as it's being sent by the nodes, we can add it back in the future should we need it.

I think @cecton wants to remove it on the substrate side as well, right? If so, is that a problem for `substrate-telemetry`?

maciejhirsz:
> > As long as it's being sent by the nodes, we can add it back in the future should we need it.
> 
> I think @cecton wants to remove it on the substrate side as well, right? If so, is that a problem for `substrate-telemetry`?

I don't think so, but it might be worth checking with `substrate-analytics` since it uses the same data as input AFAIU.

dvdplm:
That is a really good point, pretty sure it's used there. K, so @cecton it's fine to remove it here but let's keep it in the outgoing payload.
cc @mattrutherford

cecton:
I confirm there is indeed a ts field in substrate-analytics: https://github.com/paritytech/substrate-analytics/blob/master/src/web/root.rs#L251

But... the payload is completely different in V2. Would it still work with substrate-analytics if substrate nodes send a completely different payload?

mattrutherford:
The timestamp is used in substrate-analytics and I think it could be useful in other cases too, so please if possible let's keep it in the substrate telemetry messages. If the data structure is different I can prepare a PR for that, no problem

cecton:
OK I tested on current master and on my new branch and it works.

dvdplm (approved):
LGTM. 

It'd be good to have a test or two to prove backwards compatibility.

cecton:
> It'd be good to have a test or two to prove backwards compatibility.

I think there were no test at all for this. I will add some for V1 and V2 then.

maciejhirsz (approved):
Looks good to me, a simple test with `serde_json::from_str` should suffice.

cecton:
(I will wait a bit for the substrate part to be validated before merging in case there is a big change)

PR #314, opened 2021-03-25
Cap number of nodes allowed to connect for third party networks

dvdplm:
In an attempt to stem the flood of nodes connecting to the telemetry server from third party networks this PR caps the maximum number of nodes to 500.﻿


backend/src/main.rs
```diff
@@ -127,7 +127,8 @@ async fn health(aggregator: web::Data<Addr<Aggregator>>) -> Result<HttpResponse,
 /// This can be changed using the `PORT` and `BIND` ENV variables.
 #[actix_web::main]
 async fn main() -> std::io::Result<()> {
-    SimpleLogger::new().with_level(log::LevelFilter::Info).init().expect("Must be able to start a logger");
+    // SimpleLogger::new().with_level(log::LevelFilter::Info).init().expect("Must be able to start a logger");
```
dvdplm, line 130: TODO: Port over the logging changes from the other branch

backend/src/aggregator.rs
```diff
@@ -20,9 +21,27 @@ pub struct ChainEntry {
...
 
+lazy_static! {
+    /// Labels of chains we consider "first party". These chains are allowed any
+    /// number of nodes to connect.
+    static ref FIRST_PARTY_NETWORKS: HashSet<&'static str> = {
+        let mut set = HashSet::new();
+        set.insert("Polkadot");
+        set.insert("Kusama");
+        set.insert("Westend");
+        set.insert("Rococo");
```
dvdplm, line 38: This needs some vetting. Is that the full list of networks we want to consider "first party"?

lovelaced: Looks fine to me, at least; we also run Canvas but I think we can ignore that for now. The only one I'm not super sure about is "Local Testnet" but that's well under 500 at the moment so...

backend/src/aggregator.rs
```diff
@@ -39,31 +58,28 @@ impl Aggregator {
...
 
                 let addr = ctx.address();
+                let max_nodes = max_nodes(label);
                 let label: Label = label.into();
                 let cid = self.chains.add_with(|cid| {
                     ChainEntry {
                         addr: Chain::new(cid, addr, label.clone()).start(),
                         label: label.clone(),
-                        network_id: network.clone(),
+                        network_id: None, // TODO: this doesn't seem to be used anywhere. Can it be removed?
```
TarikGul, line 75: Is `network_id` still used in the `fn handler` inside of the `impl Handler`?

PR #322, opened 2021-03-26
Mute denied nodes

dvdplm:
When a node is not allowed to connect, send a `Mute` message to `NodeConnector`.

A node can be disallowed to connect if the chain is over-quota or if the whole chain is banned, i.e. on the denylist.

Also dials down the logging of latest finalized block from `info` to `debug`.

### DIFF
--- a/backend/src/aggregator.rs
+++ b/backend/src/aggregator.rs
@@ -3,7 +3,7 @@ use actix::prelude::*;
 use actix_web_actors::ws::{CloseReason, CloseCode};
 use lazy_static::lazy_static;
 
-use crate::node::connector::{Initialize, Mute};
+use crate::node::connector::{Mute, NodeConnector};
 use crate::feed::connector::{FeedConnector, Connected, FeedId};
 use crate::util::DenseMap;
 use crate::feed::{self, FeedMessageSerializer};
@@ -130,10 +130,8 @@ pub struct AddNode {
     pub node: NodeDetails,
     /// Connection id used by the node connector for multiplexing parachains
     pub conn_id: ConnId,
-    /// Recipient for the initialization message
-    pub rec: Addr<NodeConnector>,
-    /// Recipient for the mute message
-    pub mute: Recipient<Mute>,
+    /// Address of the NodeConnector actor
+    pub node_connector: Addr<NodeConnector>,
 }
 
 /// Message sent from the Chain to the Aggregator when the Chain loses all nodes
@@ -200,12 +198,12 @@ impl Handler<AddNode> for Aggregator {
     fn handle(&mut self, msg: AddNode, ctx: &mut Self::Context) {
         if self.denylist.contains(&*msg.node.chain) {
             log::warn!(target: "Aggregator::AddNode", "'{}' is on the denylist.", msg.node.chain);
-            let AddNode { mute, .. } = msg;
+            let AddNode { node_connector, .. } = msg;
             let reason = CloseReason{ code: CloseCode::Abnormal, description: Some("Denied".into()) };
-            let _ = mute.do_send(Mute { reason });
+            node_connector.do_send(Mute { reason });
             return;
         }
-        let AddNode { node, conn_id, rec } = msg;
+        let AddNode { node, conn_id, node_connector } = msg;
         log::trace!(target: "Aggregator::AddNode", "New node connected. Chain '{}'", node.chain);
 
         let cid = self.lazy_chain(&node.chain, ctx);
@@ -214,12 +212,12 @@ impl Handler<AddNode> for Aggregator {
             chain.addr.do_send(chain::AddNode {
                 node,
                 conn_id,
-                rec,
+                node_connector,
             });
         } else {
             log::warn!(target: "Aggregator::AddNode", "Chain {} is over quota ({})", chain.label, chain.max_nodes);
             let reason = CloseReason{ code: CloseCode::Again, description: Some("Overquota".into()) };
-            let _ = rec.do_send(Mute { reason });
+            node_connector.do_send(Mute { reason });
         }
     }
 }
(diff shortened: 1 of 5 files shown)


backend/src/node/connector.rs
```diff
@@ -157,6 +158,18 @@ impl NodeConnector {
     }
 }
 
+#[derive(Message)]
+#[rtype(result = "()")]
```
dvdplm, line 162: @maciejhirsz not sure what this does. Do I need it?

backend/src/node/connector.rs
```diff
@@ -157,6 +158,18 @@ impl NodeConnector {
...
+pub struct Mute;
+
+impl Handler<Mute> for NodeConnector {
+    type Result = ();
+    fn handle(&mut self, _msg: Mute, ctx: &mut Self::Context) {
+        log::trace!(target: "NodeConnector::Mute", "Muting a node");
+        ctx.stop();
```
dvdplm, line 169: When I test this locally the nodes keep trying to reconnect every 5-ish seconds so there's plenty of churn here on this. Is there a way to disconnect the socket and trigger an error on the node side? I think (but not sure) that nodes keep trying on error but not forever, so if we could do that it might be a solution to the bandwidth consumption?

maciejhirsz: I was thinking about adding `Muted` state to `ConnMultiplex` (might need a rename there) to then stop parsing messages in `StreamHandler::handle`, but still keeping the connection open. There is no way to keep it open without reading messages into buffers in actix-web AFAIK, but not doing the deserialization should help a lot (not if we are limited on the bandwidth though).

backend/src/aggregator.rs
```diff
@@ -131,6 +132,8 @@ pub struct AddNode {
     pub conn_id: ConnId,
     /// Recipient for the initialization message
     pub rec: Recipient<Initialize>,
```
maciejhirsz, line 134:
```suggestion
    pub rec: Addr<NodeConnector>,
```

And remove the `mute` recipient, no need to have two handles to the same actor.

backend/src/aggregator.rs
```diff
@@ -196,10 +199,13 @@ impl Handler<AddNode> for Aggregator {
...
         if self.denylist.contains(&*msg.node.chain) {
-            log::debug!(target: "Aggregator::AddNode", "'{}' is on the denylist.", msg.node.chain);
+            log::warn!(target: "Aggregator::AddNode", "'{}' is on the denylist.", msg.node.chain);
+            let AddNode { mute, .. } = msg;
+            let reason = CloseReason{ code: CloseCode::Abnormal, description: Some("Denied".into()) };
+            let _ = mute.do_send(Mute { reason });
             return;
         }
-        let AddNode { node, conn_id, rec } = msg;
+        let AddNode { node, conn_id, rec, mute } = msg;
```
maciejhirsz, line 208:
```suggestion
        let AddNode { node, conn_id, rec } = msg;
```

backend/src/aggregator.rs
```diff
@@ -212,6 +218,8 @@ impl Handler<AddNode> for Aggregator {
             });
         } else {
             log::warn!(target: "Aggregator::AddNode", "Chain {} is over quota ({})", chain.label, chain.max_nodes);
+            let reason = CloseReason{ code: CloseCode::Again, description: Some("Overquota".into()) };
+            let _ = mute.do_send(Mute { reason });
```
maciejhirsz, line 222:
```suggestion
            let _ = rec.do_send(Mute { reason });
```

backend/src/node/connector.rs
```diff
@@ -123,7 +121,10 @@ impl NodeConnector {
                         _ => ()
                     }
 
-                    self.aggregator.do_send(AddNode { node, conn_id, rec });
+                    let rec = ctx.address().recipient();
+                    let mute = ctx.address().recipient();
+
+                    self.aggregator.do_send(AddNode { node, conn_id, rec, mute });
```
maciejhirsz, line 127:
```suggestion
                    let rec = ctx.address();
                    self.aggregator.do_send(AddNode { node, conn_id, rec });
```

maciejhirsz (commented):
Grumbles and a suggestion how to "mute" the connector.

backend/src/node/connector.rs
dvdplm: So you're saying, instead of stopping the `NodeConnector` actor when muting a node, we'd change the state to `ConnMultiplex::Muted`? We'd still need the `conn_id`/`msg.id()` to be able to look it up in the `self.multiplex`, but maybe you're thinking of using something non-`serde` to look up that key and avoid deserializing the full `NodeMessage`?

Tell me though, is the gain you're thinking about that it's cheaper to keep the node connected but not deserialize any of their messages than closing the actor and create a new one ever `n` seconds when they reconnect?

dvdplm: I tried the following hack: I added a `muted: bool` to the `NodeConnector` actor and instead of muting a node with `ctx.close(None); ctx.stop();` I did `self.muted = true` and then changed the `StreahHandler::handle` to check this boolean as the first thing.
That would keep 1 `NodeConnector` actor around for every node but not process any of their messages. It feels pretty dirty and I'm not convinced it's better to keep the node connected rather than recreating them, but maybe you're right and that this is cheaper.

It feels to me like the *right* way to do this is for `telemetry` to close the connection with a proper reason and then change substrate to behave properly:

- When a chain is denied, `telemetry` closes the connection with a `CloseCode::Policy` (1008) (or possibly `CloseCode::Abnormal` (1006)); `substrate` does not try to reconnect.
- When a chain is over quota, `telemetry` closes the connection with a `CloseCode::Again` (1013) and the node is free to try to reconnect later (maybe with exponential backoff if we want to get fancy).

That seems like the only way we have to reduce both processing resources and bandwidth. Wdyt?

PR #331, opened 2021-03-29
listen on external interfaces in docker container

lumir-mrkva:
by default backed is only listening on localhost by default, this means that backend is unavailable to other docker containers and cannot be exposed if you attach the container to some network (in docker swarm by example).


chevdor:
Hello @lumir-mrkva, sorry it was not reviewed earlier. The same fix was proposed in #336 and I agree this will make it more comvenient. I am closing this PR since #336 has already some approvals and I would like to thank you for the PR and the nice description.

PR #335, opened 2021-04-22
Bucket nodes by genesis hash

### LINKED ISSUE #263: Filter by genesis hash
The telemetry should filter by genesis hash for known networks like `Kusama`, `Westend`, `Polkadot`, etc.

We have restarted `Westend` quite some time ago, but some nodes still produce blocks on the old network. These nodes should not be visible on the `Westend` network tab. We could use the genesis hash to ensure that nodes are reporting indeed for the correct network.

maciejhirsz:
Closes #263.

Reading `genesis_hash` from the `system.connected` message and bucketing.

Also some cleanup and rust-fmt.

### DIFF
--- a/backend/src/aggregator.rs
+++ b/backend/src/aggregator.rs
@@ -11,7 +11,7 @@ use crate::types::{ConnId, NodeDetails, NodeId};
 use crate::util::{DenseMap, Hash};
 
 pub struct Aggregator {
-    hashes: HashMap<Hash, ChainId>,
+    genesis_hashes: HashMap<Hash, ChainId>,
     labels: HashMap<Label, ChainId>,
     chains: DenseMap<ChainEntry>,
     feeds: DenseMap<Addr<FeedConnector>>,
@@ -51,7 +51,7 @@ const THIRD_PARTY_NETWORKS_MAX_NODES: usize = 500;
 impl Aggregator {
     pub fn new(denylist: HashSet<String>) -> Self {
         Aggregator {
-            hashes: HashMap::new(),
+            genesis_hashes: HashMap::new(),
             labels: HashMap::new(),
             chains: DenseMap::new(),
             feeds: DenseMap::new(),
@@ -68,7 +68,7 @@ impl Aggregator {
         label: &str,
         ctx: &mut <Self as Actor>::Context,
     ) -> ChainId {
-        let cid = match self.hashes.get(&genesis_hash).copied() {
+        let cid = match self.genesis_hashes.get(&genesis_hash).copied() {
             Some(cid) => cid,
             None => {
                 self.serializer.push(feed::AddedChain(&label, 1));
@@ -85,7 +85,7 @@ impl Aggregator {
                 });
 
                 self.labels.insert(label, cid);
-                self.hashes.insert(genesis_hash, cid);
+                self.genesis_hashes.insert(genesis_hash, cid);
 
                 self.broadcast();
 
@@ -240,7 +240,7 @@ impl Handler<DropChain> for Aggregator {
 
         if let Some(entry) = self.chains.remove(cid) {
             let label = &entry.label;
-            self.hashes.remove(&entry.genesis_hash);
+            self.genesis_hashes.remove(&entry.genesis_hash);
             self.labels.remove(label);
             self.serializer.push(feed::RemovedChain(label));
             log::info!("Dropped chain [{}] from the aggregator", label);
--- a/backend/src/aggregator.rs
+++ b/backend/src/aggregator.rs
@@ -1,14 +1,14 @@
-use std::collections::{HashMap, HashSet};
 use actix::prelude::*;
-use actix_web_actors::ws::{CloseReason, CloseCode};
+use actix_web_actors::ws::{CloseCode, CloseReason};
 use ctor::ctor;
+use std::collections::{HashMap, HashSet};
 
-use crate::node::connector::{Mute, NodeConnector};
-use crate::feed::connector::{FeedConnector, Connected, FeedId};
-use crate::util::DenseMap;
+use crate::chain::{self, Chain, ChainId, GetNodeNetworkState, Label};
+use crate::feed::connector::{Connected, FeedConnector, FeedId};
 use crate::feed::{self, FeedMessageSerializer};
-use crate::chain::{self, Chain, ChainId, Label, GetNodeNetworkState};
+use crate::node::connector::{Mute, NodeConnector};
 use crate::types::{ConnId, NodeDetails, NodeId};
+use crate::util::DenseMap;
 
 pub struct Aggregator {
     labels: HashMap<Label, ChainId>,
@@ -17,7 +17,7 @@ pub struct Aggregator {
     feeds: DenseMap<Addr<FeedConnector>>,
     serializer: FeedMessageSerializer,
     /// Denylist for networks we do not want to allow connecting.
-    denylist: HashSet<String>
+    denylist: HashSet<String>,
 }
 
 pub struct ChainEntry {
@@ -59,11 +59,7 @@ impl Aggregator {
 
     /// Get an address to the chain actor by name. If the address is not found,
     /// or the address is disconnected (actor dropped), create a new one.
-    pub fn lazy_chain(
-        &mut self,
-        label: &str,
-        ctx: &mut <Self as Actor>::Context,
-    ) -> ChainId {
+    pub fn lazy_chain(&mut self, label: &str, ctx: &mut <Self as Actor>::Context) -> ChainId {
         let cid = match self.get_chain_id(label, None.as_ref()) {
             Some(cid) => cid,
             None => {
@@ -98,7 +94,10 @@ impl Aggregator {
         let networks = &self.networks;
 
         if let Some(network) = network {
-            networks.get(&**network).or_else(|| labels.get(label)).copied()
+            networks
+                .get(&**network)
+                .or_else(|| labels.get(label))
+                .copied()
         } else {
             labels.get(label).copied()
         }
@@ -106,7 +105,9 @@ impl Aggregator {
 
     fn get_chain(&mut self, label: &str) -> Option<&mut ChainEntry> {
         let chains = &mut self.chains;
-        self.labels.get(label).and_then(move |&cid| chains.get_mut(cid))
+        self.labels
+            .get(label)
+            .and_then(move |&cid| chains.get_mut(cid))
     }
 
     fn broadcast(&mut self) {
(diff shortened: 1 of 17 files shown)


backend/src/feed/connector.rs
```diff
@@ -83,20 +83,21 @@ impl FeedConnector {
...
+                            Ok(true) => (),
+                            // Chain not found, reset hash
+                            _ => actor.chain_hash = 0,
+                        }
+                        async {}.into_actor(actor)
+                    })
+                    .wait(ctx);
             }
             "send-finality" => {
                 self.aggregator.do_send(SendFinality {
```
dvdplm, line 103: Question unrelated to this PR (and really about line 115 below): why are we doing our own ping/pong thing rather than relying on the builtin ping pong mechanism?

backend/src/util/hash.rs
```diff
@@ -0,0 +1,57 @@
+use std::fmt::{self, Debug};
+
+use serde::de::{self, Deserialize, Deserializer, Unexpected, Visitor};
+
+const HASH_BYTES: usize = 32;
+
+/// Newtype warpper for 32-byte hash values, implementing readable `Debug` and `serde::Deserialize`.
```
dvdplm, line 7: "readable", how so?

backend/src/util/hash.rs
```diff
@@ -0,0 +1,57 @@
...
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
+pub struct Hash([u8; HASH_BYTES]);
```
dvdplm, line 9: Merits a comment to our future selves: "We could use `primitive_types::H256` here, but opted for a custom type to aboid more dependencies."

backend/src/util/hash.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("hexidecimal string of 32 bytes beginning with 0x")
+    }
+
+    fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+    where
+        E: de::Error,
+    {
+        if !value.starts_with("0x") {
```
dvdplm, line 24: This is essentially the reason for not using the derive, right? Q: why not use `deserialize_with`?

backend/src/aggregator.rs
```diff
@@ -1,29 +1,29 @@
...
+use crate::chain::{self, Chain, ChainId, GetNodeNetworkState, Label};
+use crate::feed::connector::{Connected, FeedConnector, FeedId};
 use crate::feed::{self, FeedMessageSerializer};
-use crate::chain::{self, Chain, ChainId, Label, GetNodeNetworkState};
+use crate::node::connector::{Mute, NodeConnector};
 use crate::types::{ConnId, NodeDetails, NodeId};
+use crate::util::{DenseMap, Hash};
 
 pub struct Aggregator {
+    hashes: HashMap<Hash, ChainId>,
```
dvdplm, line 14: Not a great name. These are genesis hashes for chains, so I think I'd prefer the full `genesis_hashes`. 

All of these members could use some documentation to help future devs.

backend/src/aggregator.rs
```diff
@@ -1,29 +1,29 @@
...
     feeds: DenseMap<Addr<FeedConnector>>,
     serializer: FeedMessageSerializer,
     /// Denylist for networks we do not want to allow connecting.
-    denylist: HashSet<String>
+    denylist: HashSet<String>,
 }
 
 pub struct ChainEntry {
     addr: Addr<Chain>,
+    hash: Hash,
```
dvdplm, line 25: Like above, not a fan of the naming here.

backend/src/aggregator.rs
```diff
@@ -128,6 +116,8 @@ impl Actor for Aggregator {
 pub struct AddNode {
     /// Details of the node being added to the aggregator
     pub node: NodeDetails,
+    /// Genesis `Hash` of the chain the node is being added to.
```
dvdplm, line 119: Using ```[`Hash`]``` would help navigation in rustdocs.

Suggestion: "Genesis `Hash` of the chain of the node."

backend/src/feed/connector.rs
maciejhirsz: I've put that in to measure client latency, though that's not displayed anywhere other than via `console.log` in the FE currently.

backend/src/aggregator.rs
niklasad1: this is genesis hash too?

backend/src/chain.rs
```diff
@@ -457,7 +508,10 @@ impl Handler<Subscribe> for Chain {
...
             self.average_block_time,
         ));
-        self.serializer.push(feed::BestFinalized(self.finalized.height, self.finalized.hash));
+        self.serializer.push(feed::BestFinalized(
+            self.finalized.height,
+            self.finalized.hash,
+        ));
 
         for (idx, (nid, node)) in self.nodes.iter().enumerate() {
             // Send subscribtion confirmation and chain head before doing all the nodes,
```
niklasad1, line 517:
```suggestion
            // Send subscription confirmation and chain head before doing all the nodes,
```

backend/src/feed/connector.rs
```diff
@@ -83,20 +83,21 @@ impl FeedConnector {
...
+                    .send(Subscribe {
+                        chain: payload.into(),
+                        feed: ctx.address(),
+                    })
+                    .into_actor(self)
+                    .then(|res, actor, _| {
+                        match res {
+                            Ok(true) => (),
+                            // Chain not found, reset hash
+                            _ => actor.chain_hash = 0,
```
niklasad1, line 96: dq: is `chain_hash == 0 removed later/treated as an error?

backend/src/node/connector.rs
```diff
@@ -90,40 +94,46 @@ impl NodeConnector {
...
-                    let mut node = connected.node.clone();
-                    // FIXME: Use genesis hash instead of names to avoid this mess
-                    match &*node.chain {
-                        "Kusama CC3" => node.chain = "Kusama".into(),
-                        "Polkadot CC1" => node.chain = "Polkadot".into(),
-                        _ => ()
-                    }
-
-                    self.aggregator.do_send(AddNode { node, conn_id, node_connector: ctx.address() });
+                if let Payload::SystemConnected(connected) = payload {
```
niklasad1, line 124: nice :+1:

backend/src/node/message.rs
```diff
@@ -19,8 +20,8 @@ pub enum NodeMessage {
 }
 
 impl NodeMessage {
-    /// Returns a reference to the payload.
-    pub fn payload(&self) -> &Payload {
+    /// Returns the message payload.
+    pub fn into_payload(self) -> Payload {
```
niklasad1, line 24: why not `From/Into impl` instead? Might be more readable this way...

backend/src/util/hash.rs
```diff
@@ -0,0 +1,57 @@
+use std::fmt::{self, Debug};
+
+use serde::de::{self, Deserialize, Deserializer, Unexpected, Visitor};
+
+const HASH_BYTES: usize = 32;
+
+/// Newtype warpper for 32-byte hash values, implementing readable `Debug` and `serde::Deserialize`.
+#[derive(Hash, PartialEq, Eq, Clone, Copy)]
```
niklasad1, line 8:
```suggestion
#[derive(Hash, PartialEq, Eq, Clone, Copy, Debug)]
```

niklasad1 (approved):
LGTM, the naming could improved a bit as David pointed out

backend/src/feed/connector.rs
maciejhirsz: I've renamed it to `chain_label_hash`. It's been a while since I touched it (it's all just formatting), but in essence the hash is there to prevent the feed (browser client) from subscribing to the same channel twice (nothing bad would happen necessary, but it does prevent spamming the backend and getting the list of all nodes over and over again).

Setting the hash to `0` is making the check for hash equality always* fail, so that you can subscribe to a chain after first attempt failed (there were no nodes yet after backend reboot f.e.).

* as long as no chain name produces FNV hash of 0, which is astronomically unlikely.

backend/src/util/hash.rs
niklasad1: I think the `readable` means formatted as a `hex string` instead of raw bytes ^^

niklasad1, line 7:
```suggestion
/// Newtype wrapper for 32-byte hash values, implementing readable `Debug` and `serde::Deserialize`.
```

backend/src/aggregator.rs
dvdplm: @niklasad1 Yes.

backend/src/feed/connector.rs
dvdplm: > * as long as no chain name produces FNV hash of 0, which is astronomically unlikely.

When TelemetryCoin (TECO)?

backend/src/aggregator.rs
```diff
@@ -48,8 +51,8 @@ const THIRD_PARTY_NETWORKS_MAX_NODES: usize = 500;
 impl Aggregator {
     pub fn new(denylist: HashSet<String>) -> Self {
         Aggregator {
+            hashes: HashMap::new(),
```
dvdplm, line 54:
```suggestion
            genesis_hashes: HashMap::new(),
```
…or go full semantic and call this "networks"?

backend/src/aggregator.rs
```diff
@@ -1,29 +1,32 @@
...
+use crate::chain::{self, Chain, ChainId, GetNodeNetworkState, Label};
+use crate::feed::connector::{Connected, FeedConnector, FeedId};
 use crate::feed::{self, FeedMessageSerializer};
-use crate::chain::{self, Chain, ChainId, Label, GetNodeNetworkState};
+use crate::node::connector::{Mute, NodeConnector};
 use crate::types::{ConnId, NodeDetails, NodeId};
+use crate::util::{DenseMap, Hash};
 
 pub struct Aggregator {
+    hashes: HashMap<Hash, ChainId>,
```
dvdplm, line 14:
```suggestion
    genesis_hashes: HashMap<Hash, ChainId>,
```

backend/src/aggregator.rs
```diff
@@ -61,28 +64,28 @@ impl Aggregator {
     /// or the address is disconnected (actor dropped), create a new one.
     pub fn lazy_chain(
         &mut self,
+        genesis_hash: Hash,
         label: &str,
         ctx: &mut <Self as Actor>::Context,
     ) -> ChainId {
-        let cid = match self.get_chain_id(label, None.as_ref()) {
+        let cid = match self.hashes.get(&genesis_hash).copied() {
```
dvdplm, line 71:
```suggestion
        let cid = match self.genesis_hashes.get(&genesis_hash).copied() {
```

backend/src/feed/connector.rs
dvdplm: You mean like `their_pong_to_our_ping.timestamp - our_ping.timestamp`?

I wonder if surfacing that info in the lower websocket layer would be useful to have. Is it true that most websocket impls respond to `PING`/`PONG` frames right away (soketto does this)? If so it'd be a bit more accurate than doing it ourselves.

Maybe open a `nicetohave` ticket to investigate?

backend/src/aggregator.rs
maciejhirsz: Oof, I missed that when looking at code.

backend/src/aggregator.rs
maciejhirsz: I like `genesis_hashes`, I reckon it has lower gravity.

dvdplm (approved):

backend/src/util/hash.rs
```diff
@@ -0,0 +1,58 @@
+use std::fmt::{self, Debug};
+
+use serde::de::{self, Deserialize, Deserializer, Unexpected, Visitor};
+
+const HASH_BYTES: usize = 32;
+
+/// Newtype wrapper for 32-byte hash values, implementing readable `Debug` and `serde::Deserialize`.
+// We could use primitive_types::H256 here, but opted for a custom type to aboid more dependencies.
```
niklasad1, line 8:
```suggestion
// We could use primitive_types::H256 here, but opted for a custom type to avoid more dependencies.
```

maciejhirsz: Will fix that on my new branch :)

PR #336, opened 2021-04-22
Set telemetry to listen on all interfaces by default in the backend Dockerfile

dergudzon:
Added telemetry's parameters for correct launch backend in docker-compose

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -21,4 +21,4 @@ RUN apt-get -y update && apt-get -y install openssl && apt-get autoremove -y && 
 
 EXPOSE 8000
 
-ENTRYPOINT ["telemetry"]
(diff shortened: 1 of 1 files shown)


dvdplm (approved):
Those values are the default I believe, but I'm fine stating them explicitly.

chevdor:
So this change is indeed required to have the image work by default without having to pass extra parameters.

`:8000` is the default already however telemetry listens to the local interface (`127.0.0.1`) by default which makes it quite useless in a docker context. The readme suggests passing `-l 0.0.0.0:8000` but indeed we make it simpler for users by having this the default.

chevdor:
This is the same than #331. I will keep this one open since we have already some approvals.

maciejhirsz (approved):

PR #337, opened 2021-04-29
Preparing backend to receive data from shards

maciejhirsz:
+ Moved `backend` to `backend/core` (could use a better name)
+ Added `/shard_submit/{chain_hash}` endpoint that can receive `bincode` encoded binary message payloads from shards.


backend/Cargo.toml
```diff
@@ -1,30 +1,10 @@
...
-log = "0.4"
-simple_logger = "1.11.0"
-num-traits = "0.2"
-parking_lot = "0.11"
-reqwest = { version = "0.11.1", features = ["blocking", "json"] }
-rustc-hash = "1.1.0"
-clap = "3.0.0-beta.2"
-ctor = "0.1.20"
+[profile.dev]
+opt-level = 3
```
dvdplm, line 7: Why this change?

backend/core/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "telemetry"
+version = "0.3.0"
+authors = ["Parity Technologies Ltd. <admin@parity.io>"]
+edition = "2018"
+license = "GPL-3.0"
+
+[dependencies]
+actix = "0.11.1"
+actix-web = { version = "4.0.0-beta.4", default-features = false }
```
dvdplm, line 10: beta 6 is out, maybe time to upgrade?

backend/core/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+actix-http = "3.0.0-beta.4"
+bincode = "1.3.3"
+bytes = "1.0.1"
+chrono = { version = "0.4", features = ["serde"] }
+fnv = "1.0.7"
+hex = "0.4.3"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = { version = "1.0", features = ["raw_value"] }
+thiserror = "1.0.24"
+primitive-types = { version = "0.9.0", features = ["serde"] }
```
dvdplm, line 21: If this is a dependency now maybe we can use `H256` for the genesis hash instead of our own type?

backend/core/src/main.rs
```diff
@@ -103,6 +105,27 @@ async fn node_route(
...
+    let genesis_hash = hash_str.parse()?;
+
+    let mut res = ws::handshake(&req)?;
+
+    let aggregator = aggregator.get_ref().clone();
+
+    Ok(res.streaming(ws::WebsocketContext::with_codec(
+        ShardConnector::new(aggregator, genesis_hash),
+        stream,
+        Codec::new().max_size(10 * 1024 * 1024), // 10mb frame limit
```
dvdplm, line 125: Use the `CONT_BUF_LIMIT` constant here?

backend/core/src/shard/connector.rs
```diff
@@ -0,0 +1,156 @@
...
+                ctx.stop();
+            }
+        });
+    }
+
+    fn handle_message(&mut self, msg: ShardMessage, ctx: &mut <Self as Actor>::Context) {
+        let ShardMessage { conn_id, payload } = msg;
+
+        // TODO: get `NodeId` for `ShardConnId` and proxy payload to `self.chain`.
+    }
```
dvdplm, line 83: So this is a stub essentially?

backend/core/src/shard/connector.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+        self.continue_frame(bytes);
+    }
+
+    fn continue_frame(&mut self, bytes: &[u8]) {
+        if self.contbuf.len() + bytes.len() <= CONT_BUF_LIMIT {
+            self.contbuf.extend_from_slice(&bytes);
+        } else {
+            log::error!("Continuation buffer overflow");
+            self.contbuf = BytesMut::new();
```
dvdplm, line 98: Why not `self.contbuf.clear()`?

backend/core/src/shard/connector.rs
```diff
@@ -0,0 +1,156 @@
...
+        };
+
+        match bincode::options().deserialize(&data) {
+            Ok(msg) => self.handle_message(msg, ctx),
+            #[cfg(debug)]
+            Err(err) => {
+                log::warn!("Failed to parse shard message: {}", err,)
+            }
+            #[cfg(not(debug))]
+            Err(_) => (),
```
dvdplm, line 153: This is a bit dangerous no? I imagine you don't want to spam the logs in production, but isn't it a bit risky to not know about deserialization issues? Can't we let devops take care of log ingestion (and eventually filter out unneeded logs) for us?

dvdplm (approved):
lgtm, left a few questions and suggestions but overall lgtm.

dvdplm:
@maciejhirsz is this meant to be a draft still?

backend/core/src/main.rs
```diff
@@ -103,6 +105,27 @@ async fn node_route(
     )))
 }
```
TarikGul, line 107:
```suggestion
Entry point for submitting a shard
```

backend/core/src/main.rs
TarikGul: DQ: What is the significance of the 10mb frame limit?

backend/core/src/shard/connector.rs
```diff
@@ -0,0 +1,156 @@
...
+    genesis_hash: Hash,
+    /// Chain address to which this multiplex connector is delegating messages
+    chain: Option<Addr<Chain>>,
+    /// Mapping `ShardConnId` to `NodeId`
+    nodes: DenseMap<NodeId>,
+    /// Buffer for constructing continuation messages
+    contbuf: BytesMut,
+}
+
+impl Actor for ShardConnector {
```
TarikGul, line 37: Is there an `Actor` trait coming from `use actix::prelude::*` for this impl.

backend/core/src/shard/connector.rs
```diff
@@ -0,0 +1,156 @@
...
+
+impl ShardConnector {
+    pub fn new(aggregator: Addr<Aggregator>, genesis_hash: Hash) -> Self {
+        Self {
+            hb: Instant::now(),
+            aggregator,
+            genesis_hash,
+            chain: None,
+            nodes: DenseMap::new(),
+            contbuf: BytesMut::new(),
```
TarikGul, line 61: Per the docs for `ByteMut::new()`

> Creates a new BytesMut with default capacity.
> 
> Resulting object has length 0 and unspecified capacity. This function does not allocate.

I am assuming that `allocate` just means that `BytesMut` does not use any dynamic memory handling? Does that mean everything is just on the active stack?

backend/core/src/shard/connector.rs
```diff
@@ -0,0 +1,156 @@
...
+                Item::FirstText(bytes) | Item::FirstBinary(bytes) => {
+                    self.start_frame(&bytes);
+                    return;
+                }
+                Item::Continue(bytes) => {
+                    self.continue_frame(&bytes);
+                    return;
+                }
+                Item::Last(bytes) => {
+                    self.continue_frame(&bytes);
```
TarikGul, line 135: This is just to get the last remaining bytes before stopping?

TarikGul (approved):
LGTM, it took me a while to piece together some of the crates and ideas going on, and look into the actix docs, and so on but I understand and can follow along with the logic now. Just a few question for my own sake in there to get a better understanding.

backend/core/src/shard/connector.rs
dvdplm: Behind the scenes `Bytes` and `BytesMut` are backed by a `Vec`, which uses heap memory managed by an allocator. So it **does** use dynamic memory handling. What the docs are pointing out is that `new()` does not pre-allocate a buffer for future use (if you want that, use `::with_capacity()`). Check out [the source](https://docs.rs/bytes/1.0.1/src/bytes/bytes_mut.rs.html#165-167) to see for yourself! :)

backend/core/src/shard/connector.rs
dvdplm: Indeed, it's [here](https://docs.rs/actix/0.11.1/actix/trait.Actor.html).

Notice how the trait provides default implementations for most (all?) methods, which means that people like us only has to implement whatever methods we need to override and let the defaults be.

backend/core/src/main.rs
dvdplm: Websocket communication is frame-based, i.e. the payload is split into chunks of convenient size. Here we're saying that to us "convenient" means "less than 10mb".
The [gory details](https://tools.ietf.org/html/rfc6455#section-5.2).

backend/core/src/main.rs
dvdplm: I think this endpoint is more like `/// Endpoint for shards to submit data`.

TarikGul: Updated, thanks!

backend/core/src/main.rs
TarikGul: Ahh, okay thanks for the description make sense, and those docs are gold thank you very much!

backend/core/src/shard/connector.rs
TarikGul: Hence, the `fn started`, and `fn stopped` override. That's really useful. That makes a lot of sense now for when I am looking at any `impl` and considering what traits are being imported!

backend/core/src/shard/connector.rs
TarikGul: Figured it out! Resolving.

backend/core/src/shard/connector.rs
TarikGul: TIL: You can look at the source via the docs. Incredibly cool.

 Yea that definitely answers my question! Thank you :)

dvdplm: > TIL: You can look at the source via the docs. Incredibly cool.

Yeah once you get used to the perks of rustdocs it's hard to not miss it in other languages.

maciejhirsz:
Do not delete the branch.

PR #342, opened 2021-05-26
Backend Shard

maciejhirsz:
Initial implementation of the Backend Shard that is able to connect to Backend Core, and proxy `bincode`d messages to it on a single connection.


backend/core/src/chain.rs
```diff
@@ -257,29 +255,41 @@ impl Message for GetNodeNetworkState {
...
 
-        if node_connector
-            .try_send(Initialize {
-                nid,
-                conn_id,
-                chain,
-            })
-            .is_err()
-        {
+        if source.init(nid, chain) {
```
dvdplm, line 292: I think this reads a bit backwards, `if init() then remove`. Consider hoisting out the `is_err()` check here and invert the if/else.

backend/core/src/chain.rs
```diff
@@ -257,29 +255,41 @@ impl Message for GetNodeNetworkState {
...
+                false
+            }
+        }
+    }
+}
+
 impl Handler<AddNode> for Chain {
     type Result = ();
 
     fn handle(&mut self, msg: AddNode, ctx: &mut Self::Context) {
```
dvdplm, line 281: Maybe I missed it, but where do we check if a node was already connected `Direct`/`Shard` and then tries to open a second connection?

backend/core/src/aggregator.rs
```diff
@@ -124,10 +124,19 @@ pub struct AddNode {
     pub node: NodeDetails,
     /// Genesis [`Hash`] of the chain the node is being added to.
     pub genesis_hash: Hash,
-    /// Connection id used by the node connector for multiplexing parachains
-    pub conn_id: ConnId,
-    /// Address of the NodeConnector actor
-    pub node_connector: Addr<NodeConnector>,
+    /// Source from which this node is being added (Direct | Shard)
+    pub source: NodeSource,
```
dvdplm, line 128: dq: why not put the source in `NodeDetails`?

backend/core/src/aggregator.rs
```diff
@@ -188,25 +197,38 @@ pub struct GetNetworkState(pub Box<str>, pub NodeId);
 #[rtype(result = "usize")]
 pub struct GetHealth;
 
+impl NodeSource {
+    pub fn mute(&self, reason: CloseReason) {
```
dvdplm, line 201: Docs

backend/core/src/aggregator.rs
maciejhirsz: `NodeDetails` needs to be serializable, while `NodeSource` includes a ref to an actor.

backend/core/src/chain.rs
maciejhirsz: We don't, there is nothing stopping anyone from spawning a bunch of telemetry connections to the server (doesn't matter if they go through shards or directly).

backend/core/src/chain.rs
maciejhirsz: Yeah, I just hacked this quickly, should have left a TODO of some kind. Was thinking about returning an enum so it's clear what's happening.

backend/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 members = [
     "core",
     "shared",
```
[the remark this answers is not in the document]
maciejhirsz: Yeah, that's a good suggestion!

backend/core/src/chain.rs
dvdplm: Ok, that's probably fine as long as nobody's trying to dos the server, which is not an entirely unlikely thing to happen. Hmm.

maciejhirsz:
Killing this PR, had to rebase and squash to fix all the conflicts with reverts and changes post revert happening on master, made a new branch on `backend-shard-rebased` that fixes this and will open a new PR against master from there.

PR #345, opened 2021-06-02
Remove network state from frontend

gilescope:
As agreed here ( https://github.com/paritytech/substrate/issues/8999 ) removing network state from being visible in Telemetry front end as first step in removing it. 
(Next step will be stopping the publishing of the information to the telemetry server.)


dvdplm (approved):

maciejhirsz (approved):
LGTM

PR #338, opened 2021-05-03
Backend shared utils

maciejhirsz:
Extracted out from `backend/core`, to be used in `backend/shard` in future PRs.

Requires #337 to be merged to master first (this PR will be changed to target `master` instead of `mh-backend-core` when that happens).

PR #348, opened 2021-06-08
Backend Shards

maciejhirsz:
This combines #338 and #342 into one PR with squashed history since we've made a mess on master with reverts.


jsdw:
Closing this as work is progressing in `jsdw-sharding` :)

PR #349, opened 2021-06-21
Sharded Telemetry Server

jsdw:
An async hyper+soketto+tokio (-actix) telemetry server, split into:
- a `telemetry_core` binary which the UI connects to and receives information from to display, and shards connect to to relay information to it.
- a `telemetry_shard` binary which takes JSON feeds from nodes and sends them to the `telemetry` process to be aggregated and sent to UI feeds.

The telemetry core process works by spinning up a single aggregator loop which receives messages from shards and feeds, and updates the node state or subscribes/unsubscribes feeds as necessary.

At some point:
- Allow shards to connect to multiple Telemetry Cores, not just the one. This would allow us to scale the core service too when more feeds connect by spinning up more of them.
- And/or allow a telemetry core to spin up multiple aggregator loops and distribute connected feeds across them
- Allow aggregators to only handle messages relating to certain genesis hashes (chains) to split the shard side of the work across aggregators rather than broadcasting it to all of them.

I'll leave `master` alone for now (to avoid any automated deployment bits), and merge this into long lived `sharded` branch which can accomodate other changes (benchmark/e2e test tooling and any deployment config) before we merge it to master.

# Manual testing

To try this thing out locally, you can do the following:

## In the substrate repository:

Start up an "alice" node in one terminal and a bob node in another (pointed to where shard 1 will start).

```bash
cargo run -- \
  --tmp \
  --chain local \
  --alice \
  --port 30333 \
  --ws-port 9944 \
  --rpc-port 9933 \
  --node-key 0000000000000000000000000000000000000000000000000000000000000001 \
  --validator \
  --telemetry-url 'ws://localhost:8001/submit 1'
```

```bash
cargo run -- \
  --tmp \
  --chain local \
  --bob \
  --port 30334 \
  --ws-port 9945 \
  --rpc-port 9934 \
  --validator \
  --bootnodes /ip4/127.0.0.1/tcp/30333/p2p/12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp \
  --telemetry-url 'ws://localhost:8001/submit 1'
```

Now, let's start two more nodes on a different chain and pointed to a different telemetry shard:

```bash
cargo run -- \
  --tmp \
  --chain dev \
  --alice \
  --port 30335 \
  --ws-port 9946 \
  --rpc-port 9935 \
  --node-key 0000000000000000000000000000000000000000000000000000000000000001 \
  --validator \
  --telemetry-url 'ws://localhost:8002/submit 1' \
  --name AliceDev
```

```bash
cargo run -- \
  --tmp \
  --chain dev \
  --bob \
  --port 30336 \
  --ws-port 9947 \
  --rpc-port 9936 \
  --validator \
  --bootnodes /ip4/127.0.0.1/tcp/30335/p2p/12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp \
  --telemetry-url 'ws://localhost:8002/submit 1' \
  --name BobDev
```
## In this repository+branch

On another three terminals (sorry about all the terminals..), run the following from within the `backend` folder of this repo:

The main telemetry process that UIs will connect to:

```
cargo run --bin telemetry
```

A couple of shards to receive telemetry and forward it on:

```
cargo run --bin shard -- -l 127.0.0.1:8001
```

```
cargo run --bin shard -- -l 127.0.0.1:8002
```

And finally, on _yet another_ terminal and in t he `frontend` folder of this repo, we'll start the UI up so that we can see what's going on:

```
yarn start
```

Now, visit http://localhost:3000 and watch the data start coming in as nodes connect to the started shards.

Have a go at killing shards, or the telemetry core, or nodes and see what happens.


backend/common/src/util/null.rs
```diff
@@ -0,0 +1,136 @@
+use serde::de::{Deserialize, Deserializer, IgnoredAny};
+use serde::ser::{Serialize, Serializer};
+
+/// Alternative to `serde::de::IgnoreAny` that implements `Serialize`.
+/// Will serialize to `null` in JSON, or empty data in bincode.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct NullAny;
```
maciejhirsz, line 7: I believe this can go away with separate json and non-json `Payload`.

backend/shard/src/aggregator.rs
```diff
@@ -0,0 +1,229 @@
...
+                        // if this fails, it probably means the connection has died already anyway.
+                        let _ = closer.send(());
+                    }
+
+                    // We've told everything to disconnect. Now, reset our state:
+                    close_connections = vec![];
+                    to_local_id.clear();
+                    muted.clear();
+
+                    connected_to_telemetry_core = true;
```
maciejhirsz, line 141: It looks to me like `connected_to_telemetry_core` should be more of an enum like:

```rust
enum State {
    Connected {
        close_connections: ...,
        to_local_id: ...,
        muted: ...,
    },
    Disconnected,
}
```
So that whoever touches the code later cannot introduce a bug that puts the aggregator into an invalid state (e.g.: read or write to `muted` without checking for connection status first).

jsdw: Ooh, I like that, thanks!

backend/common/src/util/null.rs
jsdw: I think so too; there are a couple of things I think I'll be able to remove when I tidy up the common things :)

backend/shard/src/aggregator.rs
jsdw: Alas it's not quite that straightforward, because I want to hold onto the `close_connections` while we are disconnected, and call them when we reconnect. But I'll keep this in mind; something very similar might work well!

.github/workflows/backend.yml
```diff
@@ -14,21 +14,30 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Build
+
+    - name: Build telemetry executables (in debug mode)
       working-directory: ./backend
-      run: cargo build --verbose
+      run: cargo build --bins --verbose
```
niklasad1, line 20: cool, didn't know about `--bins` TIL

backend/common/Cargo.toml
```diff
@@ -1,29 +1,21 @@
 [package]
-name = "telemetry"
-version = "0.3.0"
+name = "common"
+version = "0.1.0"
 authors = ["Parity Technologies Ltd. <admin@parity.io>"]
 edition = "2018"
-license = "GPL-3.0"
```
niklasad1, line 6: no license?

jsdw: Ah whoops; I regenerated the file so must have forgot to put that back; thanks for spotting!

backend/common/src/log_level.rs
```diff
@@ -0,0 +1,37 @@
+/// A log level that can be decoded from a string (and so can be used
```
niklasad1, line 1: I don't get this, can't you just use https://docs.rs/log/0.4.14/log/enum.LevelFilter.html#impl-FromStr instead of this type?

jsdw: Good question! I don't see why not offhand :)

backend/shard/src/connection.rs
```diff
@@ -0,0 +1,195 @@
...
+{
+    use bincode::Options;
+    use soketto::handshake::{Client, ServerResponse};
+
+    let host = telemetry_uri.host().unwrap_or("127.0.0.1");
+    let port = telemetry_uri.port_u16().unwrap_or(8000);
+    let path = telemetry_uri.path();
+
+    let socket = TcpStream::connect((host, port)).await?;
+    socket.set_nodelay(true).unwrap();
```
niklasad1, line 110: return err?

backend/shard/src/connection.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    // Open a websocket connection with the relemetry core:
```
niklasad1, line 112:
```suggestion
    // Open a websocket connection with the telemetry core:
```

backend/shard/src/connection.rs
```diff
@@ -0,0 +1,195 @@
...
+                    // Log the error but otherwise ignore it and keep running:
+                    log::warn!("Failed to decode message from Backend Core: {:?}", err);
+                }
+            }
+        }
+    });
+
+    // This task receives messages from the aggregator,
+    // encodes them and sends them to the telemetry core:
+    let (tx_to_connection, mut rx_from_aggregator) = mpsc::channel(10);
```
niklasad1, line 166: what's the rationale behind setting channel size to 10?

backend/shard/src/json_message/hash.rs
```diff
@@ -0,0 +1,214 @@
...
+impl Display for Hash {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("0x")?;
+
+        let mut ascii = [0; HASH_BYTES * 2];
+
+        hex::encode_to_slice(self.0, &mut ascii)
+            .expect("Encoding 32 bytes into 64 bytes of ascii; qed");
+
+        f.write_str(std::str::from_utf8(&ascii).expect("ASCII hex encoded bytes canot fail; qed"))
```
niklasad1, line 131:
```suggestion
        f.write_str(std::str::from_utf8(&ascii).expect("ASCII hex encoded bytes can't fail; qed"))
```

backend/shard/src/connection.rs
jsdw: Nothing very tangible; I think I read that the default actix channel size was 10 (for the previous telemetry impl) so I thought I'd start there (though I also think we primarily ignored that size by using `do_send` to send messages).

Really I need to benchmark this and figure out what the channel size should be or whether some channels should be unbounded (but I was keen to allow for some backpressure for now)

backend/shard/src/connection.rs
jsdw: Good point!

backend/shard/src/connection.rs
niklasad1: ok, cool we were bitten quite badly about this in [jsonrpsee previously with a deadlock w.r.t bounded channels](https://github.com/paritytech/jsonrpsee/issues/143) that's the reason I'm asking :)

jsdw: Thanks for the pointer! I have pondered disconnecting nodes if the shard can't keep up (that may also allow the next node connection to be load balanced onto a shard that's under less stress). Currently, outgoing connections to feeds are unbounded (but I can def see a case for doing something smarter here and dropping messages if the buffer grows too much; I don't like the idea that a feed can be deliberately slow and cause issues that way!

Definitely more thought is needed in this area!

backend/telemetry_shard/src/main.rs
```diff
@@ -0,0 +1,190 @@
...
+                    Ok(msg) => msg,
+                };
+
+                // Close message? Break to close connection.
+                if msg.is_close() {
+                    break;
+                }
+
+                // If the message isn't something we want to handle, just ignore it.
+                // This includes system messages like "pings" and such, so don't log anything.
```
dvdplm, line 146: In soketto such messages are not forwarded to consuming code I think.

jsdw: The docs suggest that pings are responded to (and I "handle" close messages above already), but since you can also check `msg.is_ping()` I thought it best to be safe and ignore any such messages if they are seen here despite being handled.

backend/telemetry_shard/src/aggregator.rs
```diff
@@ -0,0 +1,271 @@
...
+                    FromWebsocket::Add {
+                        message_id,
+                        ip,
+                        node,
+                        genesis_hash,
+                    },
+                ) => {
+                    // Don't bother doing anything else if we're disconnected, since we'll force the
+                    // node to reconnect anyway when the backend does:
+                    if !connected_to_telemetry_core {
```
dvdplm, line 176: This is probably a terrible idea, but did you consider having a separate loop waiting only for `ConnectedToTelemetryCore` so we can avoid doing even this check?

backend/telemetry_shard/src/aggregator.rs
```diff
@@ -0,0 +1,271 @@
...
+                    if muted.contains(&local_id) {
+                        continue;
+                    }
+
+                    // Send the message to the telemetry core with this local ID:
+                    let _ = tx_to_telemetry_core
+                        .send(FromShardAggregator::UpdateNode { local_id, payload })
+                        .await;
+                }
+                ToAggregator::FromWebsocket(disconnected_conn_id, FromWebsocket::Disconnected) => {
```
dvdplm, line 221: Wouldn't we get this message from all nodes when there is a network issue and the connection from the shard to the world goes down?

backend/shard/src/connection.rs
jsdw: (I went with `expect` in the end as I'm not expecting it to reasonably fail (and if it does it'll fail fast at least))

backend/telemetry_shard/src/connection.rs
```diff
@@ -0,0 +1,195 @@
...
+use tokio_util::compat::TokioAsyncReadCompatExt;
+
+#[derive(Clone, Debug)]
+pub enum Message<Out> {
+    Connected,
+    Disconnected,
+    Data(Out),
+}
+
+/// Connect to a websocket server, retrying the connection if we're disconnected.
```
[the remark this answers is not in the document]
jsdw: It's probably specific enough to be given a more specific name, so I shall do that (it serializes/deserializes to/from bincode, but is otherwise quite generic).

backend/telemetry_shard/src/connection.rs
```diff
@@ -0,0 +1,195 @@
...
+
+            // Tell the aggregator that we're disconnected so that, if we like, we can discard
+            // messages without doing any futher processing on them.
+            if connected {
+                connected = false;
+                let _ = tx_to_external.send(Message::Disconnected).await;
+            }
+
+            // Wait a little before trying to reconnect.
+            tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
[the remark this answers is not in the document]
jsdw: I don't _think_ I'm worried about that; there will only be a small number of shards. The main thing is when nodes reestablish connections to the shards, and that should be pretty random anyway I think (based on when they are started).

If we did want to spread out the connections more, I might be tempted to introduce a random delay (from maybe 0-5 seconds) at the start of the feed loop, but I'd hope that it wouldn't be an issue (I think they will be spread out over ~5 secondsish at the mo).

backend/telemetry_shard/src/connection.rs
```diff
@@ -0,0 +1,195 @@
...
+#[derive(Clone, Debug)]
+pub enum Message<Out> {
+    Connected,
+    Disconnected,
+    Data(Out),
+}
+
+/// Connect to a websocket server, retrying the connection if we're disconnected.
+/// - Sends messages when disconnected, reconnected or data received from the connection.
+/// - Returns a channel that allows you to send messages to the connection.
```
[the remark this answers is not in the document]
jsdw: I added a note pointing to `common::internal_messages`, which documents all of the internal messages at least.

backend/telemetry_shard/src/connection.rs
```diff
@@ -0,0 +1,195 @@
...
+where
+    S: Sink<Message<Out>, Error = E> + Unpin + Send + Clone + 'static,
+    E: std::fmt::Debug + std::fmt::Display + Send + 'static,
+    In: serde::Serialize + Send + 'static,
+    Out: serde::de::DeserializeOwned + Send + 'static,
+{
+    // Set up a proxy channel to relay messages to the telemetry core, and return one end of it.
+    // Once a connection to the backend is established, we pass messages along to it. If the connection
+    // fails, we
+    let (tx_to_connection_proxy, mut rx_from_external_proxy) = mpsc::channel(10);
```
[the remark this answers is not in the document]
jsdw: I find naming lots of things hard! Here, I went for "proxy" to signify that they are indirect channels. The idea is that if the actual connection (and channels associated with that) goes down, these proxy channels remain happy and intact and are directed to new actual-channels when reconnection happens.

I'm very open to any better naming suggestions though; "proxy" was the closest short analogy that came to mind!

backend/telemetry_shard/src/aggregator.rs
```diff
@@ -0,0 +1,271 @@
...
+        payload: node_message::Payload,
+    },
+    /// Make a note when the node disconnects.
+    Disconnected,
+}
+
+pub type FromAggregator = internal_messages::FromShardAggregator;
+
+#[derive(Clone)]
+pub struct Aggregator(Arc<AggregatorInternal>);
```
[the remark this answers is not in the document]
jsdw: I hada quick play and `#[warn(missing_docs)]` doesn't seem to be awfully helpful here (I assume becasue it's a binary project and so there isn't a public interface that needs documenting)

backend/telemetry_shard/src/aggregator.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        // Return a handle to our aggregator:
+        Ok(Aggregator(Arc::new(AggregatorInternal {
+            conn_id: AtomicU64::new(1),
+            tx_to_aggregator,
+        })))
+    }
+
+    // This is spawned into a separate task and handles any messages coming
+    // in to the aggregator. If nobody is tolding the tx side of the channel
```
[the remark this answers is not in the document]
jsdw: I clearly need to get a spellcheck plugin :D

backend/telemetry_shard/src/aggregator.rs
```diff
@@ -0,0 +1,271 @@
...
+        mut rx_from_external: mpsc::Receiver<ToAggregator>,
+        mut tx_to_telemetry_core: mpsc::Sender<FromAggregator>,
+    ) {
+        use internal_messages::{FromShardAggregator, FromTelemetryCore};
+
+        // Just as an optimisation, we can keep track of whether we're connected to the backend
+        // or not, and ignore incoming messages while we aren't.
+        let mut connected_to_telemetry_core = false;
+
+        // A list of close channels for the current connections. Send an empty tuple to
```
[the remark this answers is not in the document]
jsdw: Yup! I tweaked the comment to be a little more precise

backend/telemetry_shard/src/aggregator.rs
```diff
@@ -0,0 +1,271 @@
...
+        // these to ask the connections to be closed.
+        let mut close_connections: Vec<mpsc::Sender<()>> = vec![];
+
+        // Maintain mappings from the connection ID and node message ID to the "local ID" which we
+        // broadcast to the telemetry core.
+        let mut to_local_id = AssignId::new();
+
+        // Any messages coming from nodes that have been muted are ignored:
+        let mut muted: HashSet<ShardNodeId> = HashSet::new();
```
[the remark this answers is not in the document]
jsdw: I'm not immediately aware of one off the top of my head, although it'd be easy to wrap a collection with some sort of bounding logic.

I had a check and made sure to clean up the muted set when a connection goes away (I wasn't), and made sure to clean up the close_connections list properly (it's now a HashMap).

A node could still cause an issue potentially by broadcasting a large number of message IDs along with a muted chain name, leading to the muted list to keep growing (but probably worse, leading to various other things also growing to account for each node ID).

backend/telemetry_shard/src/aggregator.rs
jsdw: I removed one of these checks (for messages coming form the telemetry core, because clearly it's connected if they come in), so actually there are only 2 now (for add/update nodes). I can't think of a clean way to add a separate bit of handling to pull this out offhand, and it feels like a per-message-type decision to make anyway.

backend/telemetry_shard/src/aggregator.rs
jsdw: We would indeed! It would cause a bit of a spike to handle them for sure, but I'm not sure how much of an issue that would be yet (the thing that could be an issue is sending `RemovedNode` messages off to the telemetry core, which is currently a bounded channel and perhaps should become an unbounded/metered one).

backend/telemetry_shard/src/json_message/hash.rs
```diff
@@ -0,0 +1,214 @@
+use serde::de::{self, Deserialize, Deserializer, SeqAccess, Unexpected, Visitor};
+use serde::ser::{Serialize, Serializer};
+use std::fmt::{self, Debug, Display};
+use std::str::FromStr;
+
+const HASH_BYTES: usize = 32;
+
+/// Newtype wrapper for 32-byte hash values, implementing readable `Debug` and `serde::Deserialize`.
+/// This can deserialize from a JSON string or array.
```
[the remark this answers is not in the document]
jsdw: (this was copied over from the previous version; I think the point of this is that it can read hashes from hex strings as well as from arrays of bytes, and present hashes in such a way when debugging)

backend/telemetry_shard/src/json_message/hash.rs
```diff
@@ -0,0 +1,214 @@
+use serde::de::{self, Deserialize, Deserializer, SeqAccess, Unexpected, Visitor};
+use serde::ser::{Serialize, Serializer};
+use std::fmt::{self, Debug, Display};
+use std::str::FromStr;
+
+const HASH_BYTES: usize = 32;
```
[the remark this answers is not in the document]
jsdw: Ah interesting; I think we assume 32 byte hashes in quite a lot of places and would need to sue a Vec or something if we want to support dynamic hash sizes (unless they are only a little dynamic ie a few different possibilities?).

I'll add a comment noting this; not sure whether it's worth doing anything else at the moment?

backend/telemetry_shard/src/aggregator.rs
dvdplm: Fair.

backend/telemetry_shard/src/connection.rs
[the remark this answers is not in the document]
dvdplm: I see. Hmm. Yeah, tricky to find a good name for indeed. At some point when the dust is settled it'd be good to have a graphical map of which channels send/receive from where... (or imagine an automated tool for that, that would be so cool).

backend/telemetry_shard/src/aggregator.rs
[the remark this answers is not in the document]
dvdplm: > (I assume becasue it's a binary project and so there isn't a public interface that needs documenting)

No it works in a new project:

```rust
#[warn(missing_docs)]

pub struct Thing{}

fn main() {
    println!("Hello, world!");
}
```

Both `warn` and `deny`.

backend/telemetry_shard/src/aggregator.rs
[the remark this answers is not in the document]
dvdplm: Yeah, a pesky issue. :/

backend/telemetry_shard/src/json_message/hash.rs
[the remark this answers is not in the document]
dvdplm: Cool, yeah, let's add that!

backend/telemetry_shard/src/json_message/hash.rs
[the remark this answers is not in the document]
dvdplm: > I'll add a comment noting this; not sure whether it's worth doing anything else at the moment?

That works, this is not an immediate concern.

backend/telemetry_shard/src/json_message/hash.rs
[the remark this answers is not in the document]
jsdw: Done :)

backend/telemetry_shard/src/aggregator.rs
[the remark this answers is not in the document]
jsdw: One thing I was wondering last night; how many different message IDs do we ever expect to get from a node? I'd assume that we should only ever see a few, so we could always disconnect nodes when we see more than some number to prevent a single node from leading to a lot of state accumulating.

@maciejhirsz do you have any thoughts on this?

backend/telemetry_shard/src/aggregator.rs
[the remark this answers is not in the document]
jsdw: Aah, OK so it works in the same way as on libraries actually; if a module isn't publically exported from `main.rs` then the attribute doesn't warn (unless you add it to each file).

```
#![warn(missing_docs)]

pub struct Thing{}

fn main() {
    let _ = does_warn::Wibble;
    println!("Hello, world! {:?}", doesnt_warn::Bar);
}

pub mod does_warn {
    pub struct Wibble;
}

mod doesnt_warn {
    #[derive(Debug)]
    pub struct Bar;
}
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,68 @@
...
+  CONTAINER_REGISTRY:     "docker.io/parity"
+  CONTAINER_REPO:         "substrate-telemetry-backend"
+  KUBE_NAMESPACE:         "substrate-telemetry"
+  IMAGE_FULL_NAME:        "${CONTAINER_REGISTRY}/${CONTAINER_REPO}:${CI_COMMIT_SHORT_SHA}"
+
+stages:
+  - dockerize
+  - staging
+
+#.delete_deployment:       &delete_deployment
```
niklasad1, line 11: remove?!

README.md
```diff
@@ -20,92 +20,130 @@ nvm install stable
...
+
+   ```
+   docker run --rm -it --network=telemetry \
+       --name frontend \
+       -p 3000:80 \
+       -e SUBSTRATE_TELEMETRY_URL=ws://localhost:8000/feed \
+       substrate-telemetry-frontend
+   ```
+
+   **NOTE:** Here we used `SUBSTRATE_TELEMETRY_URL=ws://localhost:8000/feed`. This will work if you test with everything running locally on your machine but NOT if your backend runs on a remote server. Keep in mind that the frontend docker image is serving a static site running your browser. The `SUBSTRATE_TELEMETRY_URL` is the WebSocket url that your browser will use to reach the backend. Say your backend runs on a remore server at `192.168.0.100`, you will need to set the IP/url accordingly in `SUBSTRATE_TELEMETRY_URL`.
```
niklasad1, line 132: note sure if this is correct because it's a little weird to run `remote server` at a local IP address but I guess it just an example

```suggestion
   **NOTE:** Here we used `SUBSTRATE_TELEMETRY_URL=ws://localhost:8000/feed`. This will work if you test with everything running locally on your machine but NOT if your backend runs on a remote server. Keep in mind that the frontend docker image is serving a static site running your browser. The `SUBSTRATE_TELEMETRY_URL` is the WebSocket url that your browser will use to reach the backend. Say your backend runs on a remote server at `192.168.0.100`, you will need to set the IP/url accordingly in `SUBSTRATE_TELEMETRY_URL`.
```

backend/README.md
```diff
@@ -0,0 +1,19 @@
...
+- [telemetry_shard](./telemetry_shard): A Shard. It's expected that multiple of these will run. Nodes will connect to Shard instances and send JSON telemetry to them, and Shard instances will each connect to the Telemetry Core and relay on relevant data to it.
+- [common](./common): common code shared between the telemetry shard and core
+- [test_utils](./test_utils): Test utilities, primarily focused around making it easy to run end-to-end tests.
+- [docs](./docs): Material supporting the documentation lives here
+
+# Architecture
+
+As we move to a sharded version of this telemetry server, this set of architecture diagrams may be useful in helping to understand the current setup (middle diagram), previous setup (first diagram) and possible future setup if we need to scale further (last diagram):
+
+![Architecture Diagram](./docs/architecture.svg)
```
niklasad1, line 15: :+1:

backend/common/src/assign_id.rs
```diff
@@ -0,0 +1,84 @@
...
+        Self {
+            current_id: 0,
+            mapping: BiMap::new(),
+            _id_type: std::marker::PhantomData,
+        }
+    }
+
+    pub fn assign_id(&mut self, details: Details) -> Id {
+        let this_id = self.current_id;
+        self.current_id += 1;
```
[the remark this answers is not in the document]
jsdw: Very much so! Even if we assigned every connected node an ID, and 10k nodes connected per second, it would take ~58494241 years to hit the limit :)

(this is only used in a shard to assign IDs to the nodes that connect to it)

backend/common/src/byte_size.rs
```diff
@@ -0,0 +1,107 @@
...
+use anyhow::{anyhow, Error};
+
+#[derive(Copy, Clone, Debug)]
+pub struct ByteSize(usize);
+
+impl ByteSize {
+    pub fn new(bytes: usize) -> ByteSize {
+        ByteSize(bytes)
+    }
+    pub fn into_bytes(self) -> usize {
```
niklasad1, line 26: I find `into_bytes()` a little confusing given how it's used in the Rust ecosystem.

When I see `into_bytes` I assume you will return a `vec/slice/array` of the actual bytes.

I think `len` or `size` would be more clear in this context.

README.md
jsdw: Yeah, `192.168.0.100` isn't the best example address really 😄

backend/common/src/byte_size.rs
jsdw: Yeah, `into_bytes` definitely isn't the best choice of name (nor is `ByteSize` really)! `byte_size` works :)

backend/common/src/dense_map.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    pub fn into_iter(self) -> impl Iterator<Item = (Id, T)> {
+        self.items
+            .into_iter()
+            .enumerate()
+            .filter_map(|(id, item)| Some((id.into(), item?)))
+    }
+
+    pub fn len(&self) -> usize {
+        self.items.len() - self.retired.len()
```
niklasad1, line 119: is `items` expected to always be bigger than `retired`? Might panic

maybe saturating_sub?

backend/common/src/dense_map.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        old
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = (Id, &T)> + '_ {
+        self.items
+            .iter()
+            .enumerate()
+            .filter_map(|(id, item)| Some((id.into(), item.as_ref()?)))
```
niklasad1, line 101: dq: what happens if item is `None` here i.e, this bail the entire iteration or just return `None` and continue with the next elem?

backend/common/src/either_sink.rs
```diff
@@ -0,0 +1,57 @@
...
+    #[project = EitherSinkInner]
+    pub enum EitherSink<A, B> {
+        A { #[pin] inner: A },
+        B { #[pin] inner: B }
+    }
+}
+
+/// A simple enum that delegates implementation to one of
+/// the two possible sinks contained within.
+impl <A, B> EitherSink<A, B> {
```
niklasad1, line 14: formatting seems a bit of here; I think we are using `rustfmt` in this repo right?

but maybe not enforced by CI I guess.

backend/common/src/http_utils.rs
```diff
@@ -0,0 +1,172 @@
...
+    let accept_key = generate_websocket_accept_key(key.as_bytes(), &mut accept_key_buf);
+
+    // Tell the client that we accept the upgrade-to-WS request:
+    let response = Response::builder()
+        .status(hyper::StatusCode::SWITCHING_PROTOCOLS)
+        .header(hyper::header::CONNECTION, "upgrade")
+        .header(hyper::header::UPGRADE, "websocket")
+        .header("Sec-WebSocket-Accept", accept_key)
+        .body(Body::empty())
+        .expect("bug: failed to build response");
```
niklasad1, line 90: I think this is infallible?!

backend/common/src/http_utils.rs
```diff
@@ -0,0 +1,172 @@
...
+/// Check if there is a header of the given name containing the wanted value.
+fn header_contains_value(
+    headers: &hyper::HeaderMap,
+    header: hyper::header::HeaderName,
+    value: &[u8],
+) -> bool {
+    pub fn trim(x: &[u8]) -> &[u8] {
+        let from = match x.iter().position(|x| !x.is_ascii_whitespace()) {
+            Some(i) => i,
+            None => return &x[0..0],
```
niklasad1, line 156:
```suggestion
            None => return &[],
```

backend/common/src/most_seen.rs
```diff
@@ -0,0 +1,247 @@
...
+/// seen the most is.
+#[derive(Debug)]
+pub struct MostSeen<T> {
+    current_best: T,
+    current_count: usize,
+    others: HashMap<T, usize>,
+}
+
+impl<T: Default> Default for MostSeen<T> {
+    fn default() -> Self {
```
niklasad1, line 30: hrm, can you explain the difference between `new()` and `default()` such that why is `current_count: 0` in default and `current_count: 1` in new?

backend/common/src/node_types.rs
```diff
@@ -1,16 +1,33 @@
...
-
-pub type NodeId = usize;
-pub type ConnId = u64;
 pub type BlockNumber = u64;
 pub type Timestamp = u64;
-pub type Address = Box<str>;
 pub use primitive_types::H256 as BlockHash;
 
-#[derive(Deserialize, Debug, Clone)]
+///
```
niklasad1, line 29: no docs?

backend/common/src/node_types.rs
```diff
@@ -21,43 +38,70 @@ pub struct NodeDetails {
     pub startup_time: Option<Box<str>>,
 }
 
-#[derive(Deserialize, Debug, Clone, Copy, PartialEq, Eq, Default)]
+///
```
niklasad1, line 41: no docs?

backend/common/src/node_types.rs
```diff
@@ -21,43 +38,70 @@ pub struct NodeDetails {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 pub struct NodeStats {
     pub peers: u64,
     pub txcount: u64,
 }
 
+impl Serialize for NodeStats {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
```
niklasad1, line 49: why manual implementation of `serialize` and `deserialize`?

backend/common/src/node_types.rs
```diff
@@ -68,48 +112,72 @@ pub struct NodeHardware {
...
 
-impl Serialize for NodeIO {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        let mut tup = serializer.serialize_tuple(1)?;
-        tup.serialize_element(self.used_state_cache_size.slice())?;
-        tup.end()
+///
```
niklasad1, line 164: docs?

backend/common/src/ws_client/connect.rs
```diff
@@ -0,0 +1,254 @@
...
+pub type RawSender = soketto::connection::Sender<tokio_util::compat::Compat<tokio::net::TcpStream>>;
+
+/// The receive side of a Soketto WebSocket connection
+pub type RawReceiver =
+    soketto::connection::Receiver<tokio_util::compat::Compat<tokio::net::TcpStream>>;
+
+/// A websocket connection. From this, we can either expose the raw connection
+/// or expose a cancel-safe interface to it.
+pub struct Connection {
+    tx: soketto::connection::Sender<tokio_util::compat::Compat<tokio::net::TcpStream>>,
```
niklasad1, line 39: use type alias `RawSender`?

backend/common/src/ws_client/connect.rs
```diff
@@ -0,0 +1,254 @@
...
+    rx: soketto::connection::Receiver<tokio_util::compat::Compat<tokio::net::TcpStream>>,
```
niklasad1, line 40: use type aslias `RawReceiver`?

backend/common/src/dense_map.rs
jsdw: Yeah, nothing can ever reduce the size of `self.items`, so it's always expected to be larger than `self.retired`. I wasn't 100% sure when I saw your comment so I've added a test to confirm this (in 8e9d679)

backend/common/src/either_sink.rs
jsdw: Yup; I'm using rustfmt but indeed I don't think CI enforces it at the mo. It does sometimes make the formatting a little "wierd" IMO but oh well :)

backend/common/src/dense_map.rs
niklasad1: Great

backend/common/src/dense_map.rs
niklasad1: yeah, super nice

backend/common/src/http_utils.rs
jsdw: It's infallible as long as the headers and such that I use are valid, I think! It's not infallible in the sense of `std::convert::Infallible` or something.

backend/common/src/ready_chunks_all.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        loop {
+            match this.stream.as_mut().poll_next(cx) {
+                // Flush all collected data if underlying stream doesn't contain
+                // more ready values
+                Poll::Pending => {
+                    return if this.items.is_empty() {
+                        Poll::Pending
+                    } else {
+                        Poll::Ready(Some(mem::replace(this.items, Vec::new())))
```
niklasad1, line 75:
```suggestion
                        Poll::Ready(Some(mem::take(this.items)))
```

backend/common/src/http_utils.rs
jsdw: Oh weird, I thought I did this a while ago when you mentioned it! I think I must have reverted it by mistake.

backend/common/src/most_seen.rs
jsdw: I should definitely add a comment about this because it confused me too at first!

`new` is given an actual item (like the name of a chain), so it logs that we've seen that item once (because we were just handed it).

`default` hasn't been given any items yet, so the "current best" that you'd get back is the default value of the type, and it's never actually been seen by anybody. So the first time you add any item, it becomes the "most seen item" with a count of 1.

backend/telemetry_core/src/aggregator/inner_loop.rs
```diff
@@ -0,0 +1,630 @@
...
+                if let Some(bytes) = feed_serializer.into_finalized() {
+                    let _ = feed_channel.unbounded_send(ToFeedWebsocket::Bytes(bytes));
+                }
+
+                // If many (eg 10k) nodes are connected, serializing all of their info takes time.
+                // So, parallelise this with Rayon, but we still send out messages for each node in order
+                // (which is helpful for the UI as it tries to maintain a sorted list of nodes). The chunk
+                // size is the max number of node info we fit into 1 message; smaller messages allow the UI
+                // to react a little faster and not have to wait for a larger update to come in. A chunk size
+                // of 64 means each mesage is ~32k.
```
niklasad1, line 441:
```suggestion
                // of 64 means each message is ~32k.
```

backend/common/src/node_types.rs
jsdw: I think I just added those to help visually separate the various types in that file better than they were.

I can def think about adding actual docs :D

backend/common/src/node_types.rs
jsdw: Another good candidate to add a comment for I think!

These types are used in feed messages sent to users, and so they came along with custom serialize impls already so that they would serialize to a more compact format. I then added custom deserialize impls so that I could deserialize these things for testing.

But yeah, in summary: we wanted a more compact format than the default serialization would have given :)

backend/telemetry_core/src/find_location.rs
```diff
@@ -0,0 +1,227 @@
...
+            client,
+            cache: Arc::new(RwLock::new(cache)),
+        }
+    }
+
+    pub async fn locate(&self, ip: Ipv4Addr) -> Result<Option<Arc<NodeLocation>>, reqwest::Error> {
+        // Return location quickly if it's cached:
+        let cached_loc = {
+            let cache_reader = self.cache.read();
+            cache_reader.get(&ip).map(|o| o.clone())
```
niklasad1, line 113:
```suggestion
            cache_reader.get(&ip).cloned()
```

backend/telemetry_core/src/main.rs
```diff
@@ -0,0 +1,468 @@
...
+    let recv_handle = tokio::spawn(async move {
+        loop {
+            let mut bytes = Vec::new();
+
+            // Receive a message, or bail if closer called. We don't care about cancel safety;
+            // if we're halfway through receiving a message, no biggie since we're closing the
+            // connection anyway.
+            let msg_info = tokio::select! {
+                msg_info = ws_recv.receive_data(&mut bytes) => msg_info,
+                _ = &mut recv_closer_rx => { break }
```
niklasad1, line 216: style grumble, I don't like the braces :P

```suggestion
                _ = &mut recv_closer_rx => break,
```

niklasad1 (approved):
Overall, the PR is very clean and looks good, maybe in the future we could "resultify" the APIs instead doing this much of unwrap.

I didn't review the logic super careful because it's already battle-tested, mostly style stuff ^

backend/common/src/node_types.rs
niklasad1: ok, got it.

backend/common/src/rolling_total.rs
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    /// Fetch the current rolling total that we've accumulated. Note that this
+    /// is based on the last seen times and values, and is not affected by the time
+    /// that it is called.
+    pub fn total(&self) -> Val {
+        self.total
+    }
+
+    /// Fetch the current time source, incase we need to modify it.
```
dvdplm, line 151: s/incase/in case/

backend/common/src/rolling_total.rs
```diff
@@ -0,0 +1,337 @@
...
+pub struct UserTimeSource(Instant);
+impl UserTimeSource {
+    pub fn new(time: Instant) -> Self {
+        UserTimeSource(time)
+    }
+    pub fn set_time(&mut self, time: Instant) {
+        self.0 = time;
+    }
+    pub fn increment_by(&mut self, duration: Duration) {
+        self.0 += duration;
```
dvdplm, line 186: Is addition to `Instant` saturating?

backend/common/src/node_types.rs
```diff
@@ -68,48 +112,72 @@ pub struct NodeHardware {
...
+        let mut tup = serializer.serialize_tuple(3)?;
+        // These are "one-way": we can't deserialize again from them to MeanLists:
+        tup.serialize_element(self.upload.slice())?;
+        tup.serialize_element(self.download.slice())?;
+        tup.serialize_element(self.chart_stamps.slice())?;
+        tup.end()
+    }
+}
+
+///
```
dvdplm, line 129: Docs (same in a few other places)

backend/telemetry_core/src/state/chain.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+/// Labels of chains we consider "first party". These chains allow any
+/// number of nodes to connect.
+static FIRST_PARTY_NETWORKS: Lazy<HashSet<&'static str>> = Lazy::new(|| {
+    let mut set = HashSet::new();
+    set.insert("Polkadot");
+    set.insert("Kusama");
+    set.insert("Westend");
+    set.insert("Rococo");
```
dvdplm, line 77: Should we add the common good parachains here as well? `statemine`/`statemint`?

backend/telemetry_core/src/state/chain.rs
```diff
@@ -0,0 +1,386 @@
...
+            finalized: Block::zero(),
+            block_times: NumStats::new(50),
+            average_block_time: None,
+            timestamp: None,
+            genesis_hash,
+        }
+    }
+
+    /// Can we add a node? If not, it's because the chain is at its quota.
+    pub fn can_add_node(&self) -> bool {
```
dvdplm, line 100: `is_overquota` might have been a good name here too?

backend/telemetry_core/src/state/chain.rs
```diff
@@ -0,0 +1,386 @@
...
+            if let Some(details) = node.update_details(now, propagation_time) {
+                feed.push(feed_message::ImportedBlock(nid.into(), details));
+            }
+        }
+    }
+
+    /// Check if the chain is stale (has not received a new best block in a while).
+    /// If so, find a new best block, ignoring any stale nodes and marking them as such.
+    fn update_stale_nodes(&mut self, now: u64, feed: &mut FeedMessageSerializer) {
+        let threshold = now - STALE_TIMEOUT;
```
dvdplm, line 287: Saturating sub here might be best?

dvdplm (approved):
Overall LGTM. Good job here, I especially like the attention paid to docs&comments.

README.md
```diff
@@ -20,92 +20,130 @@ nvm install stable
...
+
+### Terminal 4 - Node
 
 Follow up installation instructions from the [Polkadot repo](https://github.com/paritytech/polkadot)
 
+If you started the backend binaries with their default arguments, you can connect a node to the shard by running:
+
 ```sh
-polkadot --dev --telemetry-url ws://localhost:8000/submit
+polkadot --dev --telemetry-url 'ws://localhost:8001/submit 1'
```
ltfschoen, line 65: could we please mention why it expects ` 1` at the end of the url and what it means?

backend/telemetry_core/src/state/chain.rs
jsdw: If we ever hit an underflow error here we've done something very wrong, so I'd be happy to let it panic personally

backend/telemetry_core/src/state/chain.rs
jsdw: Agreed!

backend/telemetry_core/src/state/chain.rs
jsdw: I don't know enough to have an opinion on that, but I'd be happy to add them! I can't see any nodes from those on the current telemetry server yet; is that expected (I guess they aren't a thing just yet?)?

backend/common/src/rolling_total.rs
jsdw: I wouldn't assume so. This source is only actually used in tests, so if we overflow the time I think I'd be happy to chalk it up to programmer error and let it panic

backend/common/src/id_type.rs
```diff
@@ -0,0 +1,90 @@
...
+                ty.0
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    //! Mostly we're just checking that everything compiles OK
+    //! when the macro is used as expected..
```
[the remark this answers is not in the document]
jsdw: Maybe; I just sortof added them as "module" comments, but of course it doesn't benefit anybody either way really!

README.md
dvdplm: That is documented in substrate/polkadot, but sure, why not.

jsdw: I added a note about this verbosity level (I know it's a strange format, but in part I think it's because you can provide multiple `--telemetry-url` arguments, each with a different verbosity).

backend/telemetry_core/src/state/chain.rs
```diff
@@ -97,15 +97,15 @@ impl Chain {
...
+    pub fn is_overquota(&self) -> bool {
         // Dynamically determine the max nodes based on the most common
         // label so far, in case it changes to something with a different limit.
         self.nodes.len() < max_nodes(self.labels.best())
     }
 
     /// Assign a node to this chain.
     pub fn add_node(&mut self, node: Node) -> AddNodeResult {
-        if !self.can_add_node() {
+        if !self.is_overquota() {
```
dvdplm, line 108: Oops, I think you need to flip the logic around now.

jsdw: Oh, of course! I should have gone for is_within_quota maybe, or indeed, flipping the logic!

PR #351, opened 2021-07-05
Attempt to cut off feeds when they are too slow to accept messages.

jsdw:
This is an attempted approach to cut feeds off if they are too slow to accept new messages.

The intention is that we don't block on sending messages to feeds with immediate message flushing, and instead try to `flush` concurrently as part of the select loop..

AFAICT `.flush()` is cancel-safe (it ultimately runs a non-async fn that returns `Ok`, some `Err` or specifically a `WouldBlock` err), so My belief is that it's OK to use in a `select` loop in this way (it doesn't matter if it's cancelled the moment something else "wins").

That said, when using this approach, the UI seems to become flaky and less responsive. For instance, selecting a chain using the tabs along the top doesn't seem to update the selected tab as reliably. My guess is that it's dropping messages, but it's hard to tell.

Does anything stand out as an obvious issue?

(Side note: I realised I could not pick a buffer size in bytes; it's #messages instead. Less than ideal, but there should be a sensible number we can use in any case).

### DIFF
--- a/backend/telemetry/src/main.rs
+++ b/backend/telemetry/src/main.rs
@@ -217,7 +217,7 @@ where
 
 /// This handles messages coming from a feed connection
 async fn handle_feed_websocket_connection<S>(
-    mut websocket: ws::WebSocket,
+    websocket: ws::WebSocket,
     mut tx_to_aggregator: S,
 ) -> (S, ws::WebSocket)
 where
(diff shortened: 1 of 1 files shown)


dvdplm (commented):
I don't have anything intelligent to say here I'm afraid. 
Out of curiosity, what happens if you remove the first arm, `msg = ws_sink.flush(), if needs_flush => {` and instead do the flushing in the other arm:

```
msg = rx_from_aggregator.next() => {
  ws_sink.start_send_unpin(ws::Message::binary(bytes)).unwrap();
  ws_sink.flush();
}
```
If that still drops messages it'd be a hint that the problem is not in your code?

jsdw:
That is something I tried for my own sanity, and it worked (with an `.await` on the flush).

Another thing that seemed to improve my results was making the `select!` biased, so it would prefer to run the `flush` action if it was ready. I'm really not sure why that would have made a difference offhand (the action doesn't do anything significant, and I was finding this on iterations where I didn't do anything at all in the action).

Hopefully a test harness thing will shed some light on it!

Failing all of that, metering the channel to feeds and disconnecting them based on how full that is is another approach that I'd suspect would work much more reliably!

(side note: `tokio::select` feels like a higher level API offhand, but its semantics make it feel very much like a lower level primitive that needs using with great care!)

jsdw:
Closing this as an alternate approach has been implemented in `jsdw-sharding`

PR #353, opened 2021-07-09
Listen on all interfaces from the backend container by default

chevdor:
When running in Docker, the backend listens to :8000 but on 127.0.0.1 by default.
This is not very useful when using docker unless users make the unsafer choice to use their host network.

This PR allows listening on "all" interfaces (inside the container...) so that requests from the outside of the container do reach :8000 by default.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -21,4 +21,4 @@ RUN apt-get -y update && apt-get -y install openssl && apt-get autoremove -y && 
 
 EXPOSE 8000
 
-ENTRYPOINT ["telemetry"]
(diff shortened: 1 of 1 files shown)


jsdw (commented):
Thanks! It's worth noting perhaps that I have addressed this in the upcoming sharded version (see the `jsdw-sharding` branch, and the docker-compose bits are all working there now :)

(I'm not sure whether `--listen` is provided as an arg as part of the deployment process; having it set as part of the `ENTRYPOINT` may lead to a dupe arg somewhere, and perhaps using `CMD` is better?)

chevdor:
Closing this PR since you already have it in the upcoming new version.

PR #356, opened 2021-07-28
Helm

arshamparity:
Helm chart and CI file have been added to the project.


helm/values.yaml
```diff
@@ -0,0 +1,110 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount:
+  shard: 3
+  core: 1
+
+image:
+  repository: parity/substrate-telemetry-backend
+  pullPolicy: IfNotPresent
```
lovelaced, line 11: Shouldn't this be "always" if the image is meant to update on new releases? I may be misunderstanding, though.

helm/values.yaml
```diff
@@ -0,0 +1,110 @@
...
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+autoscaling:
+  enabled: false
```
lovelaced, line 86: I guess the first deployment will be without autoscaling enabled? I did notice the `hpa` helm file above.

helm/values.yaml
arshamparity: The image tag is specified explicitly as arguments for the helm command and it overrides this one in values.yaml file in a way that for each commit, we would have a unique docker image that would be used for that deployment. Setting `pullPolicy: IfNotPresent` would prevent images from being downloaded when deploying a previous commit on the cluster.

.gitlab-ci.yml
```diff
@@ -0,0 +1,68 @@
...
+#  - helm uninstall -n "$KUBE_NAMESPACE" "$KUBE_NAMESPACE"
+
+.deploy-k8s:              &deploy-k8s
+  image:                  paritytech/kubetools:3.5.3
+  script:
+    - echo "Deploying using image $IMAGE_FULL_NAME"
+    - echo "Using Helm `helm version --short`"
+    - export KUBERNETES_VERSION_TAG="$CI_PIPELINE_ID"
+    - |-
+      sed -i "s/version:.*/version: $KUBERNETES_VERSION_TAG/" ./helm/Chart.yaml
```
gabreal, line 21: i think the version of the chart shouldn't change with the kubernetes version. is `appVersion` the version of the deployed app, so it should be replaced with the version of `substrate-telemetry`?

helm/templates/hpa.yaml
```diff
@@ -0,0 +1,28 @@
...
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ include "substrate-telemetry.fullname" . }}
+  labels:
+    {{- include "substrate-telemetry.labels" . | nindent 4 }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ include "substrate-telemetry.fullname" . }}
```
gabreal, line 12: i'm not sure if i understand this correctly. should this be the name of the shard deployment i.e. `telemetry-shard`?

helm/values.yaml
gabreal: i think one should not rely on the fact that the image has been pulled before. the node pool could have been updated and then alertering the deployment could lead to issues that are harder to debug. but if you set the deployment tag explicitly it should be fine.

helm/templates/hpa.yaml
arshamparity: the hpa is created by default when you run `helm create ...`, I didn't remove it so we can use it later with some adjustments. If it's necessary, the file could be removed as well.

helm/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
jsdw, line 24: I'm not sure what this version number is (k8s?). I don't think much thought has ever gone into the version listed in the telemetry repos. Can this field just be removed?

helm/templates/core-deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+            - "telemetry_core"
+          args:
+            - "--listen"
+            - "0.0.0.0:8000"
```
jsdw, line 42: perhaps better to interpolate the port here with `{{ .Values.service.core.targetPort }}` to ensure it remains in sync?

helm/templates/shard-deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+            - "telemetry_shard"
+          args:
+            - "--listen"
+            - "0.0.0.0:8000"
```
jsdw, line 42: Interpolate the port as `{{ .Values.service.shard.targetPort }}` to keep it in sync?

helm/templates/shard-deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+            - "--core"
+            - "http://telemetry-core/shard_submit"
```
jsdw, line 44: Should this be `"http://telemetry-core:{{ .Values.service.core.port }}/shard_submit"` to keep it in sync with the exposed port of the core service?

helm/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+    {{- include "substrate-telemetry.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": test
+spec:
+  containers:
+    - name: wget
+      image: busybox
+      command: ['wget']
+      args: ['{{ include "substrate-telemetry.fullname" . }}:{{ .Values.service.port }}']
```
jsdw, line 14: Should `/health` be appended to this (`/` will 404)?

helm/values.yaml
jsdw: I've had annoying-to-debug issues in the past when not using `Always`; it feels safer to err on the side of caution?

helm/values.yaml
jsdw: The core cannot be autoscaled. Exactly 1 instance of it can be running, since it is stateful. A "next step" in the architecture could allow multiple cores, but even then the shards would need restarting (for now, unless we did something smarter) if new cores were added because they'd need to find out the new IP address to send telemetry to.

Shards could be autoscaled up, but removing a shard (while fine) would lead to a temporary loss of telemetry for all connected nodes until they reconnected to one of the still-present shards again.

jsdw (commented):
Looks good to me; just a few comments mainly around making sure the ports stay in sync with values.yaml.

Now to see if I can find out why I can't actually connect to a shard properly using a websocket connection.

helm/templates/core-deployment.yaml
PierreBesson, line 42: Will there be custom args that could be passed through here ?

jsdw: There definitely are other args that could be used, but I'm not sure how much I care about being able to pass them in via values.yaml versus having to update the helm chart to change them personally :)

jsdw (approved):
I'm happy to see this merged into the `jsdw-sharding` branch and to have any comments addressed by pushing straight to there if you prefer, just for the sake of moving and testing changes quickly.

helm/values.yaml
```diff
@@ -83,11 +86,19 @@ resources: {}
...
+    maxReplicas: 6
+    targetCPUUtilizationPercentage: 80
+    targetMemoryUtilizationPercentage: 80
+  # telemetry-core is not scalable atm.
+  core:
+    enabled: false
+    minReplicas: 3
+    maxReplicas: 6
+    targetCPUUtilizationPercentage: 80
+    targetMemoryUtilizationPercentage: 80
```
jsdw, line 101: I'm assuming that these lines don't mean anything since `enabled: false` above? Can they be removed/set to min/max or 1?

helm/values.yaml
arshamparity: added HorizontalPodAutoscaling to telemetry-shard deployment.

helm/Chart.yaml
arshamparity: It's your application version and only matters when you want to use helm to rollback to a previous version. The CI is configured in a way that it changes this version to the commit hash, so this guarantees that the appVersion is unique for each deployment and could also be used to rollback the helm chart.

helm/templates/shard-deployment.yaml
arshamparity: added

helm/templates/tests/test-connection.yaml
arshamparity: this was the default test in the helm chart, I removed it as we don't use this test. we use Kubernetes for liveness and readiness checks.

helm/templates/core-deployment.yaml
arshamparity: Generally this helm chart is tailored for this application and it doesn't mean to be a general helm chart for deploying multiple substrate-telemetry deployments. So I think adding the args (if any) to this file manually would make sense atm.

helm/values.yaml
arshamparity: they don't do anything, we don't use this variables in any helm template files, so setting `enabled:` to `true` or `false` would affect the deployment at all. they are just there for later use when substrate-core becomes scalable.

arshamparity:
The telemetry-shard service is available at http://substrate-telemetry.parity-stg.parity.io
The telemetry-core service is available at http://substrate-telemetry-core.parity-stg.parity.io

P.S: telemetry services are only accessible via VPN and GCP internal networks.

helm/templates/core-deployment.yaml
arshamparity: added

.gitlab-ci.yml
arshamparity: good catch.
remove that line.

helm/templates/shard-deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+            - "telemetry_shard"
+          args:
+            - "--listen"
+            - "0.0.0.0:{{ .Values.service.shard.targetPort }}"
+            - "--core"
+            - "http://telemetry-core:{{ .Values.service.core.targetPort }}/shard_submit"
```
jsdw, line 44:
```suggestion
            - "http://telemetry-core:{{ .Values.service.core.port }}/shard_submit"
```

(this might be my fault, but yeah, we need the service port and not the internal port :))

jsdw:
Merging to jsdw-sharding; we can continue pushing commits and working from it there  :)

PR #361, opened 2021-08-10
Frontend helm

arshamparity:
adding frontend to the helm chart and the CI

### DIFF
--- a/helm/templates/frontend-env-config.yaml
+++ b/helm/templates/frontend-env-config.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: frontend-env-config
+data:
+  env-config.js: |
+    window.process_env = {
+      SUBSTRATE_TELEMETRY_URL: "ws://feed.telemetry.parity-stg.parity.io/feed",
+      SUBSTRATE_TELEMETRY_SAMPLE: "",
+    }
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -5,12 +5,19 @@
 replicaCount:
   shard: 3
   core: 1
+  frontend: 1
 
 image:
-  repository: docker.io/parity/substrate-telemetry-backend
-  pullPolicy: IfNotPresent
-  # Overrides the image tag whose default is the chart appVersion.
-  tag: master-b4117d3a
+  backend:
+    repository: docker.io/parity/substrate-telemetry-backend
+    pullPolicy: IfNotPresent
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: master-b4117d3a
+  frontend:
+    repository: docker.io/parity/substrate-telemetry-frontend
+    pullPolicy: IfNotPresent
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: master-b4117d3a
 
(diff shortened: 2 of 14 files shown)


helm/templates/frontend-env-config.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: frontend-env-config
+data:
+  env-config.js: |
+    window.process_env = {
+      SUBSTRATE_TELEMETRY_URL: "ws://feed.telemetry.parity-stg.parity.io/feed",
```
jsdw, line 8: I guess untimately this value should be injected from the values file, so that we can configure it for live etc

helm/values.yaml
```diff
@@ -44,32 +58,71 @@ service:
...
-  hosts:
-    - host: chart-example.local
-      paths:
-        - path: /
-          pathType: ImplementationSpecific
-  tls: []
-  #  - secretName: chart-example-tls
-  #    hosts:
-  #      - chart-example.local
+  shard:
```
jsdw, line 77: It's hard for me to follow this too much without just deploying it and seeing what it looks like when deployed.

Does ingress being `enabled: false` for the shard and core mean that neither are exposed from k8s (the shard's `/submit` endpoint and core's `/feed` endpoints need to be reachable outside)?

I didn't notice anything in this PR but might have missed it: did you have any thoughts on how to *only* expose the shard's `/submit` endpoint and core's `/feed` endpoint outside of k8s and keep the other paths internal-only?

Ultimately, for me, it's probably easiest if we try deploying whatever we have and then seeing whether it works and looks sensible from there. I'm sure there will be various bits we'll want to tweak when we see how it all works!

jsdw (approved):
There's a bunch I can't really comment on without seeing it deployed and running, so I'm happy to have it merged to the `jsdw-sharding` branch if that helps get it deployed :)

I haven't tested the `docker-compose` thing yet to see whether that still works with your changes, but IMO it's not the most important thing to have working (for me, being able to start the individual things easily in dev is most important, and I don't really use the docker-compose stuff anyway)

helm/templates/frontend-env-config.yaml
arshamparity: this could also be a variable in the values.yaml file, but since we are not going to change every now and then I think keeping it like this would be fine. Otherwise we can just make a new variable and reference it here.

helm/values.yaml
arshamparity: there are two ways for exposing a service to outside of the Kubernetes cluster:
1- using ingress
2- using a dedicated load balancer

I used a dedicated load balancer for shards and core service, because I didn't want to pass the traffic through the ingress controller as it might affect other deployments in the cluster, so I decided to have a separate loadbalancer for the telemetry components for now. The disadvantage of using loadbalancers is that they don't understand layer 7 traffic as they only operate on layer 4(TCP/UDP).
for production, I would suggest creating a dedicated ingress controller for telemetry service and expose all the services using that ingress controller. this way we would be able to have layer 7 decisions like blocking or allowing traffic to certain endpoints and because this is a dedicated ingress controller, it wouldn't affect other deployments.

PR #365, opened 2021-08-12
Jsdw minor fixes

jsdw:
- Bump a few test thresholds to try and reduce e2e failures (Having run the tests over a hundred times on build-host without failures now, I'm hoping that will translate to fewer fails in CI too!)
- Add GitHub CI to enforce rustfmt and doc link checking
- Tweak GitLab CI to deploy to staging on merges to master and remove refs to jsdw-sharding branch

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -3,40 +3,46 @@ name: Backend CI
 on:
   push:
     paths:
-      - '.github/workflows/**'
+      - '.github/workflows/backend.yml'
       - 'backend/**'
       - '!frontend/**'
 
+env:
+  CARGO_TERM_COLOR: always
+
+defaults:
+  run:
+    working-directory: ./backend
+
 jobs:
   build:
+    name: Check Code
     runs-on: ubuntu-latest
-
     steps:
-    - uses: actions/checkout@v2
+      - name: Checkout sources
+        uses: actions/checkout@v2.3.4
 
-    - name: Build telemetry executables (in debug mode)
-      working-directory: ./backend
-      run: cargo build --bins --verbose
+      - name: Install Rust stable toolchain
+        uses: actions-rs/toolchain@v1.0.7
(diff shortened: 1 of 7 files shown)


niklasad1 (approved):

dvdplm (approved):

jsdw:
Argh, the last test failed! I'm guessing github CI is slower and so more sensitive to timeouts than my bare metal runs :(

jsdw:
Ok, I split the E2E tests out and run them separately, allowing continue on error (for now). I'll raise an issue to make them less brittle in CI, but for now I don't want them to give the false impression that things are broken (but I do want them to be noticed at least in these runs)

.github/workflows/backend.yml
```diff
@@ -44,3 +57,93 @@ jobs:
...
+          toolchain: stable
+          override: true
+
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v1.3.0
+        with:
+          working-directory: backend
+
+      - name: Cargo test
+        run: cargo test --verbose --jobs 1 --features e2e 'e2e'
```
niklasad1, line 149: hrm, does this only work with one job at the time?

looks quite slow 15 mins :man_shrugging:

jsdw: Yeahh, I've noticed the e2e tests running somewhat slower than the rest. I think it's because I put them behind a feature flag, and more recompilation (possibly not cache friendly) happens as a result!

(`--jobs 1` runs 1 test at a time, but I don't think that's what makes this so slow)

jsdw: (I wonder whether it would be faster if I just mark them as ignored rather than use a feature flag, come to think of it.. I might try that first)

jsdw: (Yup, doing that made it much quicker; latest run was ~2 mins)

PR #366, opened 2021-08-12
fixing helm chart issues

arshamparity:


helm/templates/core-ingress.yaml
```diff
@@ -0,0 +1,62 @@
...
+  {{- $_ := set .Values.ingress.core.annotations "kubernetes.io/ingress.class" .Values.ingress.core.className}}
+  {{- end }}
+{{- end }}
+{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+apiVersion: networking.k8s.io/v1
+{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+apiVersion: networking.k8s.io/v1beta1
+{{- else -}}
+apiVersion: extensions/v1beta1
+{{- end }}
```
jsdw, line 15: Do we need this sort of block, since we know what version of k8s we'll be running (and probably wouldn't "test" the other branches?

helm/templates/frontend-env-config.yaml
```diff
@@ -5,6 +5,6 @@ metadata:
 data:
   env-config.js: |
     window.process_env = {
-      SUBSTRATE_TELEMETRY_URL: "ws://feed.telemetry.parity-stg.parity.io/feed",
+      SUBSTRATE_TELEMETRY_URL: "wss://feed.telemetry.parity-stg.parity.io/feed",
```
jsdw, line 8: Aah, did you find a way to allow the feed host to accept WSS instead of proxying through the frontend pod then?

helm/values.yaml
```diff
@@ -90,28 +90,32 @@ ingress:
...
           - path: /
             pathType: ImplementationSpecific
-    tls: []
-    #  - secretName: chart-example-tls
-    #    hosts:
-    #      - chart-example.local
+    tls:
+      - secretName: feed.telemetry.parity-stg.parity.io
+        hosts:
+          - feed.telemetry.parity-stg.parity.io
```
jsdw, line 109: Aah, I guess this is what allows WSS to work?

jsdw (approved):
Couple of small comments/question but happy to see it merge :)

dvdplm (approved):

helm/templates/core-ingress.yaml
arshamparity: these are the defaults for the helm chart to make sure that this deployment would be compatible with all the Kubernetes versions so you can just ignore them.
about k8s version, they are configured to be upgraded automatically, so I think it would make sense to keep these configuration here.

helm/templates/frontend-env-config.yaml
arshamparity: at the moment, I'm using the ingress-controller (just a reverse proxy like nginx) to expose the feed endpoint. I believe we wouldn't receive so much traffic from the clients so it wouldn't cause problems if we pass traffic though the ingress controller. Let's keep it this way for the stg environment to see how much traffic would come though the proxy. If so much traffic passes, we should scale the ingress controller which is quite easy.

helm/templates/frontend-nginx-configmap.yaml
```diff
@@ -34,4 +26,16 @@ data:
...
+
+        server {
+          root /usr/share/nginx/html;
+          index index.html;
+          listen 8000;
+          server_name _;
+
+          location = /favicon.ico {
+            return 204;
+          }
```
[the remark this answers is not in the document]
arshamparity: sure! I just didn't like the 404 error in the browser so I added this part. feel free to change/remove these settings.

helm/values.yaml
arshamparity: yep! this is the configuration for the ingress controller, it automatically proxies the connections to this domain to the core service with a SSL certificate.

PR #367, opened 2021-08-12
Flume, Metrics and Message control when overloaded

jsdw:
- **Make use of `flume` throughout**. It seems to benchmark well against the futures channels I've been using (see https://github.com/zesterer/flume/blob/master/misc/benchmarks.png) but importantly it also exposes a `len()` fn so we can see how many messages are queued in each channel.
- **Gather and expose metrics**. The above gives us access to some useful metrics on top of what we'd already have. These are exposed in a format compatible with Prometheus (tested using a local version of prometheus).
- **Test util updates**. Remove the "worse" soak test runner, and add options to make it easier to test a few different things, and easier to scale the test runner to use more cores (which is needed for larger scale tests).

Builds on https://github.com/paritytech/substrate-telemetry/tree/jsdw-minor-fixes. Merge this after that.


jsdw:
Note to self: change base to master, and rebase/merge this after the `jsdw-minor-fixes` stuff merges

jsdw:
(force push to undo a commit I put on the wrong branch :))

backend/common/src/dense_map.rs
```diff
@@ -131,3 +131,35 @@ where
...
     }
 }
+
+#[cfg(test)]
+mod test {
+
+    use super::*;
+
+    #[test]
+    fn len_doesnt_panic_if_lots_of_retired() {
```
dvdplm, line 141: Did you mean "removed" here?

backend/common/src/ws_client/receiver.rs
```diff
@@ -15,13 +15,12 @@
...
 
 use super::on_close::OnClose;
-use futures::channel::mpsc;
 use futures::{Stream, StreamExt};
 use std::sync::Arc;
 
 /// Receive messages out of a connection
 pub struct Receiver {
-    pub(super) inner: mpsc::UnboundedReceiver<Result<RecvMessage, RecvError>>,
+    pub(super) inner: flume::r#async::RecvStream<'static, Result<RecvMessage, RecvError>>,
```
dvdplm, line 23: Eew, that is an unfortunate choice of module name.

backend/telemetry_core/src/aggregator/aggregator.rs
```diff
@@ -74,19 +85,31 @@ impl Aggregator {
         })))
     }
 
-    // This is spawned into a separate task and handles any messages coming
-    // in to the aggregator. If nobody is tolding the tx side of the channel
-    // any more, this task will gracefully end.
+    /// This is spawned into a separate task and handles any messages coming
+    /// in to the aggregator. If nobody is tolding the tx side of the channel
```
dvdplm, line 89:
```suggestion
    /// in to the aggregator. If nobody is holding the tx side of the channel
```

backend/telemetry_core/src/aggregator/inner_loop.rs
```diff
@@ -100,6 +101,30 @@ pub enum FromFeedWebsocket {
...
+    /// How many feeds are currently subscribed to something.
+    pub subscribed_feeds: usize,
+    /// How many feeds have asked for finality information, too.
+    pub subscribed_finality_feeds: usize,
+    /// How many messages are currently queued up in internal channels
+    /// waiting to be sent out to feeds.
+    pub total_messages_to_feeds: usize,
+    /// How many messages are queued waiting to be handled by this aggregator.
+    pub total_messages_to_aggregator: usize,
+    /// How many nodes are currently known about by this aggregator.
```
dvdplm, line 120:
```suggestion
    /// How many nodes are currently known to this aggregator.
```
?

backend/telemetry_core/src/aggregator/inner_loop.rs
```diff
@@ -100,6 +101,30 @@ pub enum FromFeedWebsocket {
...
+    pub connected_nodes: usize,
+    /// How many feeds are currently connected to this aggregator.
+    pub connected_feeds: usize,
```
dvdplm, line 123: dq: what is the difference between connected feeds and subscribed feeds?

backend/telemetry_core/src/aggregator/inner_loop.rs
```diff
@@ -100,6 +101,30 @@ pub enum FromFeedWebsocket {
     Disconnected,
 }
 
+/// A set of metrics returned when we ask for metrics
+#[derive(Clone, Debug, Default)]
+pub struct Metrics {
```
dvdplm, line 106: Thinking out loud here: it would be interesting to have a few histograms too, e.g. nodes and their verbosity level (to get an inkling to how much data we're receiving), message payload size in and out, maybe connection longevity too (how often do nodes stay connected).
Another data point I wouldn't mind having is the current message rate (is that what they call a "gauge"?).

backend/telemetry_core/src/main.rs
```diff
@@ -466,3 +481,59 @@ where
     // loop ended; give socket back to parent:
     (tx_to_aggregator, ws_send)
 }
+
+async fn return_prometheus_metrics(aggregator: AggregatorSet) -> Response<hyper::Body> {
+    let metrics = aggregator.latest_metrics();
+
+    // Instead of using the rust prometheus library (which is optimised around global variables updated across a codebase),
+    // we just split out the text format that prometheus expects ourselves, using whatever the latest metrics that we've
+    // captured so far from the aggregators are. See:
```
dvdplm, line 490:
```suggestion
    // we just split out the text format that prometheus expects ourselves, and use the latest metrics that we've
    // captured so far from the aggregators. See:
```

dvdplm (approved):
LGTM. Left a few nits and suggestions (likely for future work).

backend/telemetry_core/src/aggregator/aggregator_set.rs
```diff
@@ -1,36 +1,85 @@
...
+                    // Lock, update the stored metrics and drop the lock immediately.
+                    // We discard any error; if something went wrong talking to the inner loop,
+                    // it's probably a fatal error
+                    {
+                        inner.metrics.lock().unwrap()[idx] = metrics;
+                    }
+
+                    // Sleep *at least* 10 seconds. If it takes a while to get metrics back, we'll
+                    // end up waiting longer between requests.
+                    tokio::time::sleep_until(now + tokio::time::Duration::from_secs(10)).await;
```
insipx, line 74: why `sleep_until` instead of just `sleep`?

backend/telemetry_core/src/aggregator/inner_loop.rs
```diff
@@ -39,6 +37,9 @@ pub enum ToAggregator {
     FromShardWebsocket(ConnId, FromShardWebsocket),
     FromFeedWebsocket(ConnId, FromFeedWebsocket),
     FromFindLocation(NodeId, find_location::Location),
+    /// Hand back some metrics. The provided sender is expected not to block when
+    /// a message it sent into it.
```
insipx, line 41:
```suggestion
    /// a message is sent into it.
```

insipx (approved):

backend/common/src/ws_client/receiver.rs
jsdw: Yeah, I thought so! If I had to write that too much I'd probably `use` and rename it!

backend/common/src/dense_map.rs
jsdw: `retired` is referring to the list of `retired` IDs in the densemap struct :)

backend/telemetry_core/src/aggregator/inner_loop.rs
```diff
@@ -100,6 +101,30 @@ pub enum FromFeedWebsocket {
...
+/// A set of metrics returned when we ask for metrics
+#[derive(Clone, Debug, Default)]
+pub struct Metrics {
+    /// When in unix MS from epoch were these metrics obtained
+    pub timestamp_unix_ms: u64,
+    /// How many chains are feeds currently subscribed to.
+    pub chains_subscribed_to: usize,
+    /// How many feeds are currently subscribed to something.
+    pub subscribed_feeds: usize,
+    /// How many feeds have asked for finality information, too.
```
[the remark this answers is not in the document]
jsdw: What "too" was hinting at is that it's the number of feeds that are subscribed to a chain and also have asked for finality (because you can't ask for finality without being subscribed)

backend/telemetry_core/src/aggregator/inner_loop.rs
jsdw: As soon as a connection to `/feed` is established, the aggregator can see that as a connected feed. Only when the feed sends a message like "subscribe:Polkadot" does it also count as having subscribed to something!

backend/telemetry_core/src/aggregator/inner_loop.rs
jsdw: I think I return the current message rates already (I think guage is the right term.. there was a word for numbers that are always increasing and a word for values that can go up or down and I think guage was the latter!). I think we can see the bandwidth anyway from outside the process, so I wonder how useful the bytes in/out are? But that histogram sounds like a good idea; roughly breaking down the bytes in/out per node/feed could be interesting!

When this merges, perhaps we should create an issue to track any additional metrics we'd like to add?

backend/telemetry_core/src/aggregator/aggregator_set.rs
jsdw: I could have gone either way really, but I went for "aim to have 10 seconds between each update to metrics, unless it takes more than 10 seconds for an update to come through"

jsdw:
> looks good to me apart from a couple questions. Overall I've had a good experience with flume, I like how Sync/Async methods are both available on the same Sender/Receiver objects

That's good to hear! I like the flexibility it has there too, and unlike the tokio channels it also has Stream/Sink impls, and unlike futures/tokio channels you can see lengths (and on top of all of that, it's supposed to be pretty quick) :)

backend/telemetry_core/src/aggregator/aggregator_set.rs
jsdw: (To explain the reasoning a bit more; the actual logic to gather metrics is fairly quick, but potentially the message could end up in a queue for a while if we happen to be under heavy load, so if it takes eg 9seconds to get a response, I'd rather put another message into the queue 1 second later than wait another 10 seconds before queueing another message)

backend/telemetry_core/src/aggregator/inner_loop.rs
```diff
@@ -172,28 +197,86 @@ impl InnerLoop {
...
+        while let Ok(msg) = rx_from_external.recv_async().await {
+            // ignore node updates if we have too many messages to handle, in an attempt
+            // to reduce the queue length back to something reasonable, lest it get out of
+            // control and start consuming a load of memory.
+            if metered_tx.len() > max_queue_len {
+                if matches!(
+                    msg,
+                    ToAggregator::FromShardWebsocket(.., FromShardWebsocket::Update { .. })
+                ) {
+                    continue;
```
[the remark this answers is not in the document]
jsdw: Ooh yeah, that's a really good number to have! I might sneak that into this PR as I don't think it'll be that hard to gather and it would be useful to see reported!

backend/telemetry_core/src/aggregator/inner_loop.rs
dvdplm: > When this merges, perhaps we should create an issue to track any additional metrics we'd like to add?

You read my mind.

backend/common/src/dense_map.rs
jsdw: (I made the name less awful!)

backend/telemetry_core/src/aggregator/inner_loop.rs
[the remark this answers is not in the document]
dvdplm: I see. How about this: "Number of subscribed feeds that also asked for finality information."?

backend/telemetry_core/src/aggregator/inner_loop.rs
dvdplm: So one connection could subscribe to multiple chains?

backend/telemetry_core/src/aggregator/inner_loop.rs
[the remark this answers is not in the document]
dvdplm: Please do!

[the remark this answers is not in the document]
jsdw: (done!)

backend/telemetry_core/src/aggregator/inner_loop.rs
jsdw: Nope; if a connection calls "susbcribe:ChainName" it'll unsubscribe them from any previous chain they were subscribed to, so it's just one at a time.

backend/telemetry_core/src/aggregator/inner_loop.rs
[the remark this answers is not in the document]
jsdw: I like it!

backend/common/src/dense_map.rs
```diff
@@ -138,7 +138,7 @@ mod test {
     use super::*;
 
     #[test]
-    fn len_doesnt_panic_if_lots_of_retired() {
+    fn len_doesnt_panic_if_lots_of_ids_are_retired() {
```
dvdplm, line 141: ❤️

backend/telemetry_core/src/aggregator/inner_loop.rs
dvdplm: Got it.

dvdplm:
Last [changes](https://github.com/paritytech/substrate-telemetry/pull/367/commits/b842c7fc8bbfa5b25cd087283efbd6a62db55790) still lgtm! Merge at will.

Issue #357, opened 2021-07-30
Document what verbosity is and does


chevdor:
-[ ] error here: https://github.com/paritytech/substrate-telemetry#terminal-3---node

jsdw:
With the recent merge of the sharding docs, I added a line of help in this area already ("Note: The "0" at the end of the URL is a verbosity level, and not part of the URL itself. Verbosity levels range from 0-9, with 0 denoting the lowest verbosity."). Is that sufficient to have this issue closed?

chevdor:
Yep that totally works for me. I would add a comment (if still relevant) that both url and verbosity need to be quoted together as I saw that as a common issue. ie: `--telemetry-url <url> <verbosity>` vs `--telemetry-url '<url> <verbosity>'`.
If you could link the new doc/commit with the fix, we could close that right away.

jsdw:
Closed by #368 :)

Issue #334, opened 2021-04-06
env-config.js not found


NukeManDan:
The top level `index.html` has [a reference](https://github.com/paritytech/substrate-telemetry/blob/089352c73eb1ac9103486e801b2f6bced86df4ac/frontend/public/index.html#L9) to `env-config.js` that is also in the `.gitignore` list (as it should be). 

But this leads to an error, as there is no prompt to generate this file, or (better) check to see if it exists and warn/ignore it if so. This is in the console on startup:

```
Uncaught SyntaxError: expected expression, got '<'
```

chevdor:
Hey @NukeManDan,
I would have to double check the following but the file you mention may be a generated file allowing to pass ENV.

chevdor:
See https://github.com/paritytech/substrate-telemetry/blob/master/frontend/env.sh

chevdor:
This is [where](https://github.com/paritytech/substrate-telemetry/pull/240/files#diff-ea60ef29f6f537c1c83468c00b60de28f86e06edfe4a3d91274723c6f298fdb8R31) the env-config.js is generated

The background is that we wanted a static site (smaller, simpler, etc...) but that means obviously no server and nothing to load the env. When using node in dev mode, this is taken care of for you, when using docker, there is a script that generated this js file that the static HTML can load to get the ENV from docker.

If you run directy from the console, you should be in node dev mode. Not sure what the context is but we can have a chat :)

jsdw:
It's worth noting that some of the way this works is changing as a result of recent work to add Helm charts and such for deployment of this to our internal services. We'll have to review the Dockerfiles and steps now I think to make sure that it can still be deployed using Docker/docker-compose independently, as it might be that the README needs a tweak to accomodate these changes.

chevdor:
@jsdw noted. Make sure you keep the feature that allows passing ENV variables to the UI (see [also this comment](https://github.com/paritytech/substrate-telemetry/pull/240#issuecomment-606096510)), that will be useful in K8S as well. This feature is also useful for internal networks.

chevdor:
@NukeManDan could you share the exact command you used to produce this issue?

jsdw:
@chevdor yup, definitely! The changes made so far have just tweaked the frontend Dockerfile to not generate that env-config.js file, and instead it's being mounted via a configmap (which the frontend still picks up).

I haven't had a chance today, but I'll have a pass over things next week and make sure that local docker and the readme are still correct!

[...]

PR #368, opened 2021-08-14
Simplify Helm/Docker usage

jsdw:
- Re-add the `env.sh` bits in the frontend Dockerfile, and inject the env var in Helm ratehr than mounting a configmap to achieve the same (this should address comments in #334).
- Simplify the frontend nginx.conf so that it's suitable for local runs via compose or helm runs, rather than having 2 separate nginx confs around. Remove the now-unnecessary proxying bits that were useful in the old deployment.
- Add to "verbosity" explanation to note that one shell arg needs URL+verbosity (addresses #357).

I've tested the docker-compose changes on my ubuntu laptop and it all seems to work as expected. I don't have the ability to properly test the Helm stuff without it deploying (@arshamparity perhaps you can help with that?)


chevdor:
So I tested a few things and some comments go beyond the scope of this PR.
I can relocate those comments if needed.

- indeed, docker-compose seems to be working fine
- While testing the helm chart, I noticed that the current version has quite some strong requirements on an existing infra (See `affinity` for instance). I see that many sections have been commented out, IMO `affinity` can be present but should also be commented out by default
- Most defaults in `values.yaml` are really parity specifc
- I don't have a LoadBalancer in my cluster. It is quite easy to pass values so the shard service uses NodePort. Setting this as default, may allow more users to have it work right away.
- after fixing the above, almost everything started fine expect an issue with the readiness probe of the frontend: `Readiness probe failed: Get "http://10.244.0.184:8000/": dial tcp 10.244.0.184:8000: connect: connection refused` and the frontend is stuck in a crash loop. This issue is due to the fact that I switched to a NodePort but the template only considers `targetPort` so the 2 probes are testing on the wrong port. A quick fix is to adjust the `targetPort` that makes it a bit hacky
- IMO `value.yml` should work for a "neutral" cluster such as a brand new minikube. We could have a `value-parity.yml` in the repo for our infra but that may not be the ideal location for such a file
- In the same line I think the frontend could be exposed as `NodePort` by default

After fixing the above, I could install the chart. However, my http requests to `frontend` and `shard` failed from outside the cluster. A curl from inside the cluster from a test pod did work though.

So here is my test setup:
- comment out  the `affinity` section
- change the service values as shown below

```
service:
  shard:
    type: NodePort
    port: 80
    targetPort: 80
    nodePort: 30666
    annotations:
      external-dns.alpha.kubernetes.io/hostname: submit.telemetry.parity-stg.parity.io.
  core:
    type: ClusterIP
    port: 80
    targetPort: 8000
    annotations: {}
    #  cloud.google.com/load-balancer-type: Internal
    #  networking.gke.io/internal-load-balancer-allow-global-access: "true"
    #  external-dns.alpha.kubernetes.io/hostname: feed.telemetry.parity-stg.parity.io.
  frontend:
    type: NodePort
    port: 80
    targetPort: 80
    nodePort: 30667
    annotations: {}
```

chevdor:
The issue is the `frontend-service.yaml` template in relation with my *NodePort requirement*.

We can do it more generic but a simple fix from:
```
ports:
    - port: {{ .Values.service.frontend.port }}
      targetPort: {{ .Values.service.frontend.targetPort }}
```
to
```
ports:
    - port: {{ .Values.service.frontend.port }}
      {{- if .Values.service.frontend.targetPort }}
      targetPort: {{ .Values.service.frontend.targetPort }}
      {{- end }}
      {{- if .Values.service.frontend.nodePort }}
      nodePort: {{ .Values.service.frontend.nodePort }}
      {{- end }}
```
allows me to reach the frontend.

jsdw:
@chevdor Thankyou for your thorough feedback; that's great! I don't have enough of an opinion/experience in Helm/k8s to give detailed feedback offhand myself.

Could I propose that we push to merge this current PR if it seems like an improvement (happy to be told if it's not though :D), and open a separate issue containing your feedback, which feels like it could lead to some useful work around tidying up and improving the Helm stuff now that we've got something in place?

chevdor:
I agree. Let me bring the setup up again to double check the changes of this PR and I will provide a review solely on this PR so we can close it and move on the other topic in another issue as you mentioned.

frontend/Dockerfile
```diff
@@ -15,18 +15,18 @@ FROM docker.io/nginx:stable-alpine
...
 
+# When we run the built docker image, the default command uses this
+# env var to tell the UI where to look to get feed info.
 ENV SUBSTRATE_TELEMETRY_URL=
 
 WORKDIR /usr/share/nginx/html
 
-#COPY --from=builder /opt/builder/env.sh /usr/bin/
-#RUN apk add --no-cache bash; chmod +x /usr/bin/env.sh
+COPY --from=builder /opt/builder/env.sh /usr/bin/
```
arshamparity, line 24: Instead of running `env.sh` inside the container, we can just put the output of the script in a file and mount it inside the container. users must change this file (instead of the env variable) if they're going to make some changes.

Generally, the best practice is to avoid passing config files and scripts in the build step as changing the runtime configuration needs rebuilding the image. It's not also compatible with the Kubernetes logic where we have configmaps, secrets and env variables.

I totally agree that maintaining two duplicate nginx configuration (one for docker-compose and the other for the helm chart) is not ideal, but we can't have multiple deployments on different platforms with a single configuration file. We could have a separate docker-compose directory and put some basic configuration files there for docker-compose users.

frontend/Dockerfile
```diff
@@ -15,18 +15,18 @@ FROM docker.io/nginx:stable-alpine
...
+RUN apk add --no-cache bash; chmod +x /usr/bin/env.sh
 
-#COPY --from=builder /opt/builder/nginx/nginx.conf /etc/nginx/nginx.conf
-#COPY --from=builder /opt/builder/nginx/default /etc/nginx/sites-available/default
+COPY --from=builder /opt/builder/nginx/nginx.conf /etc/nginx/nginx.conf
```
arshamparity, line 27: the above comment applies to this line as well.

frontend/Dockerfile
```diff
@@ -15,18 +15,18 @@ FROM docker.io/nginx:stable-alpine
...
 COPY --from=builder /opt/builder/build /usr/share/nginx/html
 
-EXPOSE 80
+EXPOSE 8000
```
arshamparity, line 30: I can't remember why I changed the port here, but it really make sense to use a port higher than 1024 for the application inside the container. 👍

helm/templates/frontend-deployment.yaml
```diff
@@ -54,24 +54,8 @@ spec:
             httpGet:
               path: /
               port: http
-          volumeMounts:
```
arshamparity, line 57: as I noted earlier, these changes should be reverted.

helm/values.yaml
```diff
@@ -26,7 +26,10 @@ fullnameOverride: ""
 envVars:
   shard: {}
   core: {}
-  frontend: {}
```
arshamparity, line 29: If we mount the output of the `env.sh` script inside the container instead of generating it in the runtime, we wouldn't need this variable anymore.

frontend/Dockerfile
chevdor: I will open a new issue with those k8s related topics as discussed with @jsdw. 

> Instead of running env.sh inside the container, we can just put the output of the script in a file and mount it inside the container. users must change this file (instead of the env variable) if they're going to make some changes.

I would recommend against that. That means users would have to write and tweak a JS file for the solely purpose of passing ENV vars... Not only this is a garanty for errors but that also does not make much sense.

> Generally, the best practice is to avoid passing config files and scripts in the build step as changing the runtime configuration needs rebuilding the image. It's not also compatible with the Kubernetes logic where we have configmaps, secrets and env variables.

I agree. The env-config.js file is not generated at build time, it is generated as run time, based on the ENV the user passes. When the user uses k8s, those can be passed as ConfigMaps, Secrets or ENV as you mentioned.

> I totally agree that maintaining two duplicate nginx configuration (one for docker-compose and the other for the helm chart) is not ideal, but we can't have multiple deployments on different platforms with a single configuration file. We could have a separate docker-compose directory and put some basic configuration files there for docker-compose users.

I will open the k8s related issue and we can dig into that in there.

chevdor: I created an [issue for the k8s topics](https://github.com/paritytech/substrate-telemetry/issues/370) that are not directly related to this PR.

arshamparity:
@chevdor Indeed, This helm chart is tailored for the parity infrastructure and it wasn't meant to be used for other users. Having some of those options are necessary to have the helm chart working properly with our current infrastructure. As you suggested we should have separate `values.yaml` files for public users and parity infrastructure but I don't know where would be the best location for storing the parity-specific values file.

AFAIU we would need a general helm chart for other users who want to run this deployment on their infrastructure, so most of the files in the helm chart must be changed to work properly with different values for different environments.

frontend/Dockerfile
jsdw: I think it may have been like that from before your changes. I get the impression from reading the docs that `EXPOSE` doesn't really do anything useful for us, but I wasn't very clear on it. Anyway, it may have been incorrect for a while and just not had any impact on anything :)

arshamparity: yeah! the expose option is only used when you run `docker -P` to publish all the ports.

helm/values.yaml
arshamparity: I totally agree with this as you explained earlier that this might increase the chance of mistakes by the users.

arshamparity:
@jsdw feel free to ping me if you need any help regarding the docker-compose changes (and ofc other things :-) )

frontend/Dockerfile
chevdor: Using `EXPOSE` is also good for documentation purpose and shows what ports the container exposes.

jsdw:
It's not super clear to me based on the comments so far that we've reached agreement on whether to:

a) Pass the feed URL in via an ENV var, as this PR does, or
b) Pass the feed URL via a pre-generated `env-config.js` file that is mounted into the container (as was done prior to this PR).

I'm of course in favour of (a). @arshamparity @chevdor are you guys both happy with that as well?

Is there anything else that needs changing in this PR for you guys to be happy with it?

chevdor (approved):
Putting my k8s related comments aside (moved to #370), this PR LGTM. I also tested the Frontend and once the fixes to make k8s working locally are done, using and [image built based on this PR](https://hub.docker.com/layers/chevdor/substrate-telemetry-backend/latest/images/sha256-6744bf46a66ce35c327856b078053e0eb750a77a1719825b7543cab634e27e99?context=repo), I could successfully tests and see that `window.process_env.SUBSTRATE_TELEMETRY_URL` is available. That means the `ENV` magic is working fine.

arshamparity:
Thanks @chevdor for reviewing the helm chart.
to sum up, these changes should be done:
- rolling back the helm chart changes as I'm going to change them in a separate PR (https://github.com/paritytech/substrate-telemetry/issues/370).
- mounting the Nginx configuration using the `volumes` feature of the `docker-compose` instead of copying it in the build time

jsdw:
- I didn't make any helm chart changes here in response to @chevdor's comments that were moved to that new issue; is the anything in particular in this PR that you'd prefer me to roll back?
- I tweaked the nginx config in this way so that the "plain" docker version would run in a straight forward way. To me, the nginx config is a detail that should be internal to the container (just as the telemetry backend core and shard serve content, so does the frontend pod, and that is the API that they expose). I'm not sure what the value of moving it out would be?

frontend/nginx/nginx.conf
```diff
@@ -10,19 +10,25 @@ events {
...
     keepalive_timeout  65;
     gzip  on;
     include /etc/nginx/conf.d/*.conf;
+
+    server {
+        root /usr/share/nginx/html;
+        index index.html;
+        listen 8000;
+        listen [::]:8000;
+        server_name telemetry.polkadot.io localhost;
```
arshamparity, line 32: > I tweaked the nginx config in this way so that the "plain" docker version would run in a straight forward way. To me, the nginx config is a detail that should be internal to the container (just as the telemetry backend core and shard serve content, so does the frontend pod, and that is the API that they expose). I'm not sure what the value of moving it out would be?

if you could change the `server_name telemetry.polkadot.io localhost;` to `server_name _;` it would be fine to me as well. If we are not going to change the Nginx configuration frequently, we can copy it in the build time then.

jsdw: As it happens, I just tweaked this line in my latest commit before I saw this :) Since nginx uses this block by default, I just removed the "server_name" line since I believe it's irrelevant (any host header will redirect to this block regardless of what it is)

jsdw: And that sounds good; yup, I don't expect the nginx config to need to change; it's just very basic and generic stuff to serve the static files on the port that we expose in docker :)

arshamparity (approved):

PR #376, opened 2021-08-24
gitlab-ci: changing docker repositories

arshamparity:
more details here: https://github.com/paritytech/helm-charts/issues/7


chevdor:
@arshamparity I am not sure a new repo is the best option here.
I prepared a nice explanations then lost connection and my answer as well... here it is again.

We are bringing breaking changes to telemetry, this is fine. That should be reflected in both the telemetry binaries as a major version bump, same with the docker containers.
Currently, we have a "stable" docker image tagged as "latest".

You are making a new helm chart that is planned to be working with the new version so it totally makes sense that it does not work with "latest".

I think a good alternative would be to use the same repo but tag the new docker images as "beta" (for instance). Projects such as the helm chart could point to "beta" for now, and until we promote the new telemetry to the new "official" version, it would then become "latest".

jsdw:
I've no strong opinions here, so once you guys are happy I am!

chevdor:
I have also no strong opinion either and publishing automatically is great in any case.

Facts recap. :
- we need to publish those images **somewhere** for testing/staging
- some terms being confusing... we need to publish to `<repo>/<image>:<tag>` while not stepping on the official releases 
- so we could use either an alternate `repo`, `image`, or `tag`
- I think we agree that using the same `repo`(`parity`) is easier
- the question boils down to using another `image` name or another `tag`

I just feel that another image name for the same image (just a new version) is not right.
On the other hand I think we should avoid *polluting* the main repo with tags that potentially could confuse users.

Considering all of that, I think the best trade off is to use tags with a specify pattern.

So instead of `parity/substrate-telemetry-frontend-stg:1.2.3` I would suggest we use something like `parity/substrate-telemetry-frontend:STAGING-1.2.3` or `parity/substrate-telemetry-frontend:STAGING-PR-4434`, etc....

I am open on the content of the `STAGING-1.2.3` pattern as long as:
- it is long and "scary" enough for users to type so they don't lazily default to it just because it was the latest published image :)
- it carries the fact that this is neither `latest`, `stable` nor recommended for production

arshamparity:
@chevdor I've reverted back the changes for pushing the `latest` docker tag. Now the CI tags dockers images with the "${CI_COMMIT_SHORT_SHA}-beta" variable. So as you suggested the name is long and hard enough to type or use by mistake. It's also container "beta" to inform the user that it's not `stable` nor `latest`.
I think going forward to tweak the tags based on the tags on the commits would be out of the scope of this task, so I would sugged opening an issue for CI/CD team for tweaking the CI to push the images based on the tags provided by developers (`latest`, `stable`, etc.).

arshamparity:
https://github.com/paritytech/devops/issues/980

chevdor (approved):

jsdw (approved):
Sounds good to me :)

PR #377, opened 2021-08-24
Hardening of the docker image

chevdor:
details to come


chevdor:
I am rebasing master.

frontend/nginx/nginx.conf
```diff
@@ -23,9 +22,10 @@ http {
     keepalive_timeout  65;
     gzip  on;
     include /etc/nginx/conf.d/*.conf;
+    include /etc/nginx/sites-available/*;
```
jsdw, line 25: What is this line added for?

chevdor: That looks like a left over, I will check and likely be able to remove it.

jsdw:
Everything looks good to me offhand (although I haven't tested that things still work as expected yet)! 

There is a part of my brain that's wondering whether the benefit of a read-only image is worth the slight complexity increase, so I'd be interested to know what your thoughts were around adding this?

Anyway, I'll give it a quick test tomorrow and it feels like a net win even with the added complexity, so I'm happy to approve once I've tested it!

chevdor:
> so I'd be interested to know your thoughts around why you ended up making this change

To enter a cluster running containerized polkadot nodes and telemetry, you may take the front door... or target smaller "utility" images such as telemetry to make your way in.

This PR first of all disallow running as root. Second, it prevents adding or modifying files in the container (removing files would only help disrupt the service...), thus greatly reducing the attack surface.

chevdor:
Another similar PR #379 is on its way for the backend.

arshamparity:
It looks great to me.

arshamparity (approved):

chevdor:
If that helps for the tests, I can bring a `justfile` similar to https://github.com/paritytech/substrate-api-sidecar/pull/648. It makes it easier to test.

docker-compose.yml
```diff
@@ -20,6 +20,11 @@ services:
     build:
       dockerfile: Dockerfile
       context: ./backend/
+    read_only: true
+    tmpfs:
+      - /var/cache/nginx:uid=101,gid=101
+      - /var/run:uid=101,gid=101
+      - /app/tmp:uid=101,gid=101
```
jsdw, line 27: Is this bit meant to be in the telemetry-frontend section?

chevdor: Yes, it is.
The 2 `/var` paths are to make nginx happy despite the whole container being read-only.
The `/app` is to make the frontend happy despite the whole container being read-only.

jsdw: That makes sense to me!

jsdw (approved):
LGTM once that docker-compose tmpfs is on the frontend section :)

I merged this with the backend one and forced a rebuild of images, and everything seemed to work as expected!

docker-compose.yml
chevdor: > That makes sense to me! I guess app/tmp is because we need to write the env-config file on start.

Exactly. And I made it locked, so we can write it once, and then never again. And the script doing this is read-only as well so it cannot be altered to change this behavior,

chevdor:
> LGTM once that docker-compose tmpfs is on the frontend section :)

@jsdw, indeed this is mandatory with using `--read-only`.

docker-compose.yml
jsdw: Sorry, I realise now that I wasn't clear above!

This `tmpfs` section is currently **not** being aplpied to the frontend docker image, so I believ it needs moving up a few lines to be in the correct place :)

chevdor: Oh you are totally right. I see the issue now. I will fix that.

chevdor: This is the mix up between the 2 PRs FE and BE :)
I wondered why it would even work (it should not) wihtout the tmpfs, the answer is that `read-only` also belongs to the frontend part. I will move all of that up.

Issue #378, opened 2021-08-24
Almost empty list for Moonbase Alpha


crystalin:
Looking at https://telemetry.polkadot.io/#list/Moonbase%20Alpha, there are currently 3 nodes only reported (out of the 500 nodes).
I quickly looked at the console, no error and the websocket response for `subscribe:Moonbase Alpha` only reports those nodes it seems

sea212:
[Zeitgeist telemetry information](https://telemetry.polkadot.io/#list/Zeitgeist%20Battery%20Park) is also not displayed properly. The nodes are working as expected and not outputting any error messages. Somehow I managed to get the page to show the correct list one time and the Zeitgeist tab at the top of the page also seems to show the correct number of nodes connected via telemetry, so I assume the information is available to the server but cannot be displayed for some reason.

maciejhirsz:
@crystalin @sea212 This should be fixed now, be advised the networks are now addressed by genesis hash:

Moonbase: https://telemetry.polkadot.io/#list/0x91bc6e169807aaa54802737e1c504b2577d4fafedd5a02c10293b1cd60e39527
Zeitgeist: https://telemetry.polkadot.io/#list/0xb0a8d493285c2df73290dfb7e61f870f17b41801197a149ca93654499ea3dafe

Feel free to reopen the issue if there is anything we've missed :)

crystalin:
Thank you @maciejhirsz , I confirm it works

Issue #375, opened 2021-08-24
Unable to handle networks with duplicated name


xlc:
This is a network list I just fetched from https://telemetry.polkadot.io/ (by inspecting ws messages)

You can see there are multiple `Local Testnet` and `Karura`. For some reason the frontend now decides display Karura with 3 nodes (someone's Karura private testnet), not the one with 60 nodes (the Karura mainnet).

There are some race conditions on backend & frontend.

Why the backend is not merging network with same name? It is useful to distinguish between different `Local Testnet` but there isn't any unique id exposed so impossible correctly handle them.

The frontend should also display the network with most nodes, not pick some arbitrary one.

```
[...]

Issue #350, opened 2021-06-30
Make UI subscribe by genesis hash, not chain label


jsdw:
Chain labels change and are based on whatever the most common label for a given chain that's been seen is.

The UI subscribes to chains by their label, but these labels could change at any time.

This requires the backend to maintain an uptodate mapping from chain label to chain details, so that it knows which chain a subscription is about. We run the risk of this mapping falling out of sync with reality, or of potential race conditions between a UI "subscribe to chain" message and a chain label changing (however improbabale).

I'd propose that we provide genesis hashes to the UI along with labels/node counts, and have the UI use those when subscribing to chains. This should reduce backend complexity and the scope for races.

chevdor:
That's a good idea but how do you propose handling the case of a chain fork?
Both chains would have the same genesis hash but we end up talking about 2 different chains.
Having the genesis hash as well could be usefull though.

PR #388, opened 2021-08-28
Subscribe to chains by genesis hash

maciejhirsz:
Should fix issues: #378, #375, #350.

This is a retro PR for old (currently deployed) backend, needs a companion PR for current master.

### DIFF
--- a/backend/core/src/aggregator.rs
+++ b/backend/core/src/aggregator.rs
@@ -71,13 +71,14 @@ impl Aggregator {
         let cid = match self.genesis_hashes.get(&genesis_hash).copied() {
             Some(cid) => cid,
             None => {
-                self.serializer.push(feed::AddedChain(&label, 1));
+                self.serializer
+                    .push(feed::AddedChain(&label, genesis_hash, 1));
 
                 let addr = ctx.address();
                 let max_nodes = max_nodes(label);
                 let label: Label = label.into();
                 let cid = self.chains.add_with(|cid| ChainEntry {
-                    addr: Chain::new(cid, addr, label.clone()).start(),
+                    addr: Chain::new(cid, genesis_hash, addr, label.clone()).start(),
                     genesis_hash,
                     label: label.clone(),
                     nodes: 1,
@@ -237,7 +238,7 @@ impl Handler<DropChain> for Aggregator {
             let label = &entry.label;
             self.genesis_hashes.remove(&entry.genesis_hash);
             self.labels.remove(label);
-            self.serializer.push(feed::RemovedChain(label));
+            self.serializer.push(feed::RemovedChain(entry.genesis_hash));
             log::info!("Dropped chain [{}] from the aggregator", label);
             self.broadcast();
         }
@@ -256,8 +257,9 @@ impl Handler<RenameChain> for Aggregator {
             }
 
             // Update UI
-            self.serializer.push(feed::RemovedChain(&entry.label));
-            self.serializer.push(feed::AddedChain(&new, entry.nodes));
+            self.serializer.push(feed::RemovedChain(entry.genesis_hash));
+            self.serializer
+                .push(feed::AddedChain(&new, entry.genesis_hash, entry.nodes));
 
             // Update labels -> cid map
             self.labels.remove(&entry.label);
@@ -320,12 +322,15 @@ impl Handler<Connect> for Aggregator {
 
         connector.do_send(Connected(fid));
 
-        self.serializer.push(feed::Version(31));
+        self.serializer.push(feed::Version(32));
 
         // TODO: keep track on number of nodes connected to each chain
         for (_, entry) in self.chains.iter() {
-            self.serializer
-                .push(feed::AddedChain(&entry.label, entry.nodes));
+            self.serializer.push(feed::AddedChain(
+                &entry.label,
+                entry.genesis_hash,
+                entry.nodes,
+            ));
         }
 
         if let Some(msg) = self.serializer.finalize() {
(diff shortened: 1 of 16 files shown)


backend/core/src/util/hash.rs
```diff
@@ -11,6 +12,13 @@ const HASH_BYTES: usize = 32;
 #[derive(Hash, PartialEq, Eq, Clone, Copy)]
 pub struct Hash([u8; HASH_BYTES]);
 
+impl Default for Hash {
+    #[inline(always)]
+    fn default() -> Hash {
```
niklasad1, line 17: maybe make this const instead of `default impl` such as `const fn new/zero`? Such as https://docs.rs/ethereum-types/0.12.0/ethereum_types/struct.H256.html#method.zero

niklasad1 (approved):
LGTM, clean

backend/core/src/aggregator.rs
```diff
@@ -256,8 +257,9 @@ impl Handler<RenameChain> for Aggregator {
             }
 
             // Update UI
-            self.serializer.push(feed::RemovedChain(&entry.label));
-            self.serializer.push(feed::AddedChain(&new, entry.nodes));
+            self.serializer.push(feed::RemovedChain(entry.genesis_hash));
+            self.serializer
+                .push(feed::AddedChain(&new, entry.genesis_hash, entry.nodes));
```
jsdw, line 262: I think we can just remove these lines now, since the genesis hash is identical?

jsdw:
The code changes look good to me! I haven't tested it yet though. Happy to see it merge as long as it's been tested locally to make sure that:

- Subscribing to nodes on a couple of chains still works OK,.
- Identical chain labels across a couple of chains with different genesis hashes are no longer an issue.

I'll give it a more thorough test and port the changes over to the new version when I'm back in on Tuesday if nobody else gets around to it by then :)

jsdw (approved):
I've tested (1) now, and it works at least as well as it did by the looks of it!

I did a rudimentary test of (2) using `websocat` to fake a couple of nodes, like so:

```
echo '{"id":1,"ts":"2021-07-12T10:37:47.714666+01:00","payload":{"authority":true,"chain":"Local Testnet","config":"","genesis_hash":"0x1122334455667788990011223344556677889900112233445566778899001122","implementation":"Substrate Node","msg":"system.connected","name":"Alice","network_id":"12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp","startup_time":"1625565542717","version":"2.0.0-07a1af348-aarch64-macos"}}' | websocat ws://localhost:8000/submit -n
```

The UI showed two "Local Testnet"'s when the genesis hash for 2 messages was tweaked, which indicates that it's working as expected. Node names didn't update properly, but I've seen that on the live version, and I think the UI needs a bit of a prod from other messages on resubscribing.

So, I'll tick this! Would be great if somebody was able to test (2) a little more thoroughly, but looks promising :)

backend/core/src/util/hash.rs
```diff
@@ -55,6 +63,20 @@ impl<'de> Deserialize<'de> for Hash {
...
+impl Serialize for Hash {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let mut ascii = [0u8; 2 + HASH_BYTES * 2];
+
+        ascii[..2].copy_from_slice(b"0x");
+
+        hex::encode_to_slice(&self.0, &mut ascii[2..]).expect("Encoding 32 bytes into 64 bytes of ascii; qe");
+
+        serializer.serialize_str(
+            std::str::from_utf8(&ascii).expect("Hex encoding is always valid utf8; qed"),
```
dvdplm, line 75: Maybe we can even be bold here and go with `from_utf8_unchecked()`? After all `from_utf8` + `expect` is sort of the same thing?

frontend/src/common/index.ts
```diff
@@ -9,4 +9,4 @@ import * as FeedMessage from './feed';
 export { Types, FeedMessage };
 
 // Increment this if breaking changes were made to types in `feed.ts`
-export const VERSION: Types.FeedVersion = 31 as Types.FeedVersion;
+export const VERSION: Types.FeedVersion = 32 as Types.FeedVersion;
```
dvdplm, line 12: What happens to a v31 browser connecting to a v32 feed? Will it know to reload itself? (Would be good to add to the comment what the behaviour is here)

dvdplm (approved):
lgtm, had a few questions

dvdplm:
> Node names didn't update properly, but I've seen that on the live version

Do we have an issue tracking this?

backend/core/src/util/hash.rs
jsdw: The unchecked version is `unsafe` iirc, so this one would be safer regardless :)

jsdw:
> > Node names didn't update properly, but I've seen that on the live version
> 
> Do we have an issue tracking this?

Just to clarify; I haven't seen node names fail to update at all, but there seems to be a delay in certain cases when you flick between chains, and my crappy test probably wasn't enough to trigger the update, since my "nodes" didn't send anything after the initial init message.

frontend/src/common/index.ts
maciejhirsz: @dvdplm will include a comment in the backport (frontport?) to master (and yes, it will reload itself on mismatch)

backend/core/src/aggregator.rs
maciejhirsz: The hash doesn't change, but the name can still change :).

PR #391, opened 2021-08-30
Make `Tooltip` as a function component with react hooks

CYBAI:
While trying to read the code base, I found most of the components are class component with `shouldComponentUpdate` (a.k.a `sCU`) but I wonder maybe some of them don't need the `sCU` and can be simplified as a function component.

For example, in class component based `Tooltip`, it checks `this.props.text !== nextProps.text || this.state.copied !== nextState.copied` in its `sCU`.
In React, it should already help us to trigger re-render when a prop is changed. So, we're using class components in other places, the component will always pass the same references for the callbacks (e.g. `onInit` or `onCopy`) so it should not trigger a re-render. and it should also trigger a re-render when the `text` is changed.
Also, for the `copied` state, with moving to use `React.useState`, the hooks will help us to memoize the value of the state and only trigger re-render when the state is updated.

I wonder if this improves the component and also simplify it (like with using `useRef`, we don't need to set the ref manually). Also, thanks to using function component, I found there's a redundant children in `ConsensusBlock` which might be missed to remove in [this commit](https://github.com/paritytech/substrate-telemetry/commit/ebb01c1a7db55e2bcef6059a3f3fac64f4996941#diff-108d51161fdfef38740aed57b0e10e034fb75ba6553fa071931b3eb21bb877d4L71-L76).

**Question**: Not sure if we actually need to render the children of Tooltip ☝️ , or it's just missed to be removed in [the commit](https://github.com/paritytech/substrate-telemetry/commit/ebb01c1a7db55e2bcef6059a3f3fac64f4996941#diff-108d51161fdfef38740aed57b0e10e034fb75ba6553fa071931b3eb21bb877d4L71-L76)?

Please feel free to let me know if this makes sense to you 🙏

### DIFF
--- a/frontend/src/components/Consensus/ConsensusBlock.tsx
+++ b/frontend/src/components/Consensus/ConsensusBlock.tsx
@@ -125,9 +125,8 @@ export class ConsensusBlock extends React.Component<ConsensusBlock.Props, {}> {
                   className="legend"
                   key={'block_row_' + this.props.height + '_legend'}
                 >
-                  <Tooltip text={`Block number: ${this.props.height}`}>
-                    {this.displayBlockNumber()}
-                  </Tooltip>
+                  <Tooltip text={`Block number: ${this.props.height}`} />
+                  {this.displayBlockNumber()}
                 </th>
                 <th
                   className="finalizedInfo"
--- a/frontend/src/components/Tooltip.tsx
+++ b/frontend/src/components/Tooltip.tsx
@@ -23,7 +23,6 @@ export namespace Tooltip {
   export interface Props {
     text: string;
     copy?: (cb: CopyCallback) => void;
-    className?: string;
     position?: 'left' | 'right' | 'center';
     onInit?: (update: UpdateCallback) => void;
   }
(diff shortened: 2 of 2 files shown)


frontend/src/components/Consensus/ConsensusBlock.tsx
```diff
@@ -125,9 +125,7 @@ export class ConsensusBlock extends React.Component<ConsensusBlock.Props, {}> {
                   className="legend"
                   key={'block_row_' + this.props.height + '_legend'}
                 >
-                  <Tooltip text={`Block number: ${this.props.height}`}>
-                    {this.displayBlockNumber()}
-                  </Tooltip>
+                  <Tooltip text={`Block number: ${this.props.height}`} />
```
maciejhirsz, line 128: This component is unused currently, so I guess that's why we never found this oversight :).

The child should be moved adjacent to `Tooltip`:

```suggestion
                  <Tooltip text={`Block number: ${this.props.height}`} />
                  {this.displayBlockNumber()}
```

Need to restore the method obviously :)

maciejhirsz (changes requested):
This looks good, just one clarification on the unused children in the `Tooltip`. I think I coded this defensively before going aggressive with `shouldComponentUpdate` in parent components.

The class component look less magical than hooks to me since they don't abuse hidden global state and all the lifecycle segments are clearly labeled, plus and if you just remove `shouldComponentUpdate` you end up with the same LOC count, but that might be just me because I'm an old grognard :upside_down_face:.

frontend/src/components/Consensus/ConsensusBlock.tsx
CYBAI: oh, I see! thank you for your comment!

CYBAI:
> This looks good, just one clarification on the unused children in the `Tooltip`. I think I coded this defensively before going aggressive with `shouldComponentUpdate` in parent components.
> 
> The class component look less magical than hooks to me since they don't abuse hidden global state and all the lifecycle segments are clearly labeled, plus and if you just remove `shouldComponentUpdate` you end up with the same LOC count, but that might be just me because I'm an old grognard 🙃.

Yes, to be honest, I also wondered maybe it's good to just keep it as is because it's quite easy to understand with the class component syntax.
But I also wondered that without looking at the `useEffect` one, `useState` and `useRef` are pretty neat to help us make the related parts simpler.
I will fix https://github.com/paritytech/substrate-telemetry/pull/391#discussion_r699371564 but if you think it's good to keep class component, please feel free to let me know and close this PR 🙇

jsdw:
Personally I'm a huge fan of hooks. In "one off" uses like this, you don't get to see them really shine, but when making use of them throughout an app you really get to benefit from how much more easily you can compose hook related code, and the TypeScript typing is much simpler as well often.

Hooks are basically the eventual solution to wanting "mixins" in components, and IMO they work better than the olf React mixin things ever did :)

maciejhirsz:
I never liked mixins so maybe there is that going on for me :D, either way happy to merge the PR once the one bit is fixed.

CYBAI:
Thank you so much! I've fixed it in the bbb58ee commit 🙏

jsdw (approved):
LGTM!

There are some things I don't love about this component in general (is it even possible to select and copy the text; maybe that logic can all go away? And why does the `onInit` function modify the element textContent rather than set some internal state? And can we avoid needing `onInit` at all?) but they aren't related to this PR :)

CYBAI:
> There are some things I don't love about this component in general (is it even possible to select and copy the text; maybe that logic can all go away? And why does the `onInit` function modify the element textContent rather than set some internal state? And can we avoid needing `onInit` at all?) but they aren't related to this PR :)

Yes! I would be happy to look into it! 😄  and yes, as this PR aims to move from class component to function component with hooks so I didn't make other changes in this PR 🙏

maciejhirsz (approved):

PR #394, opened 2021-09-01
Avoid using `flume::Receiver::into_stream()` to avoid memory leaks until the issue is resolved upstream

jsdw:
The changes here work around this issue:

https://github.com/zesterer/flume/issues/88

We also add `jemalloc` (for no real reason other than it was added to test memory leak things, and it's supposed to be fairly performant anyway and has some debugging hooks, so I saw no reason to remove it again).

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -438,6 +438,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "fs_extra"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2022715d62ab30faffd124d40b76f4134a550a87792276512b18d63272333394"
+
+[[package]]
 name = "funty"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -754,6 +760,27 @@ name = "itoa"
 version = "0.4.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dd25036021b0de88a0aff6b850051563c6516d0bf53f8638938edbb9de732736"
+
+[[package]]
+name = "jemalloc-sys"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d3b9f3f5c9b31aa0f5ed3260385ac205db665baa41d49bb8338008ae94ede45"
+dependencies = [
+ "cc",
+ "fs_extra",
+ "libc",
+]
+
+[[package]]
+name = "jemallocator"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43ae63fcfc45e99ab3d1b29a46782ad679e98436c3169d15a167a1108a724b69"
+dependencies = [
+ "jemalloc-sys",
+ "libc",
+]
 
 [[package]]
 name = "js-sys"
@@ -1580,6 +1607,7 @@ dependencies = [
  "hex",
  "http",
  "hyper",
+ "jemallocator",
  "log",
  "num_cpus",
  "once_cell",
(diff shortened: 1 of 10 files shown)


.gitlab-ci.yml
```diff
@@ -1,8 +1,8 @@
 variables:
   BACKEND_CONTAINER_REPO:         "docker.io/parity/substrate-telemetry-backend"
   FRONTEND_CONTAINER_REPO:        "docker.io/parity/substrate-telemetry-frontend"
-  BACKEND_IMAGE_FULL_NAME:        "${BACKEND_CONTAINER_REPO}:${CI_COMMIT_SHORT_SHA}-beta"
-  FRONTEND_IMAGE_FULL_NAME:       "${FRONTEND_CONTAINER_REPO}:${CI_COMMIT_SHORT_SHA}-beta"
+  BACKEND_IMAGE_FULL_NAME:        "${BACKEND_CONTAINER_REPO}:${CI_COMMIT_SHORT_SHA}-beta-leaktests"
```
arshamparity, line 4: I would suggest not changing the tag format as it's already tagged with `beta`.

.gitlab-ci.yml
```diff
@@ -32,13 +32,13 @@ stages:
         --create-namespace
         --namespace $KUBE_NAMESPACE
         --set image.backend.repository="${BACKEND_CONTAINER_REPO}"
-        --set image.backend.tag="${CI_COMMIT_SHORT_SHA}-beta"
+        --set image.backend.tag="${CI_COMMIT_SHORT_SHA}-beta-leaktests"
         --set image.frontend.repository="${FRONTEND_CONTAINER_REPO}"
-        --set image.frontend.tag="${CI_COMMIT_SHORT_SHA}-beta"
+        --set image.frontend.tag="${CI_COMMIT_SHORT_SHA}-beta-leaktests"
```
arshamparity, line 37: these should also be reverted

.gitlab-ci.yml
```diff
@@ -32,13 +32,13 @@ stages:
...
         --values values-parity.yaml
         $KUBE_NAMESPACE parity/substrate-telemetry
   rules:
-    - if: '$CI_COMMIT_BRANCH == "master"'
+    - if: '$CI_COMMIT_BRANCH == "jsdw-mem-testing"'
```
arshamparity, line 41: This is the only change that should happen.
I would suggest to preserve the `'$CI_COMMIT_BRANCH == "master"'` and only add ` '$CI_COMMIT_BRANCH == "jsdw-mem-testing"'` to the rules.

jsdw: Ah ok; as it stands, it doesn't seem to be running in Gitlab anyway, but happy to add the "master" back; I wasn't intending on merging this PR anywhere though (unless it's a requirement that I merge something to master to have it work at all?)

.gitlab-ci.yml
jsdw: I didn't want images being built on this branch to conflict/overwrite those that are built and deployed from master, I guess. I don't mind either way though, so happy to revert :)

arshamparity: actually, all the images would be unique anyways as their tag contains ${CI_COMMIT_SHORT_SHA} which is supposed to be unique. So the pipeline wouldn't overwrite the existing images.

.gitlab-ci.yml
arshamparity: tbh I'm not sure about your question, But I think it should be merged to the master to work properly. Maybe it's better to ask @TriplEight or @rcny .

arshamparity (approved):
looks good!

jsdw:
Now that we've found the source of the leaks (thanks, Jan!) I tidied up this PR, which had been used to test fixes, and it can now be merged into master.

It may be that `flume` is fixed and released so quickly that we don't actually need these changes, but we have the option here if we want to resolve the memory leaks on master sooner :)

maciejhirsz (approved):
LGTM!

dvdplm (approved):
nice

PR #396, opened 2021-09-03
Flume fix part 2: avoid using flume in a couple of cases, and revert attempted fix#1

jsdw:
The earlier fix that got merged to master turns out not to actually work (beyond fixing the memory leak), so this fix scraps that approach and ends up reverting some uses of flume to try and avoid the problematic polling-leads-to-memory-leak issue. We still use a flume stream in one place (because we want the metrics that it gives back), but I _believe_ that we won't have an issue with that particular usage, because we won't poll more than once before ready (unless the channel is asked to close, which will tidy up anyway).

This has been tested locally and with the e2e tests and looks like it works just fine, but getting it merged to master will be the easiest way to ensure that no memory is leaking.


backend/telemetry_core/src/main.rs
```diff
@@ -353,8 +353,7 @@ where
 {
     // unbounded channel so that slow feeds don't block aggregator progress:
     let (tx_to_feed_conn, rx_from_aggregator) = flume::unbounded();
-    let mut rx_from_aggregator_chunks =
-        ReadyChunksAll::new(common::flume_receiver_into_stream(rx_from_aggregator));
+    let mut rx_from_aggregator_chunks = ReadyChunksAll::new(rx_from_aggregator.into_stream());
```
jsdw, line 356: This is the remaining use of the potentially problematic `into_stream` method. In this case, we have a separate task which selects between items from this, and a close channel. Since the close channel only fires when we want to end, we hopefully don't `poll` the thing more than once until it's ready (suspending the task until it is).

Because the leak happens each time a stream is polled but not yet ready, my hope is that we'll be OK here.

I kept this use of the flume because we make use of metrics from these channels and I didn't want to have to remove those.

niklasad1 (approved):

backend/telemetry_core/src/main.rs
jsdw: (I reworded this into a comment in the code, since I think it's important to be aware of for now)

dvdplm (approved):
Looks sane to me.

PR #399, opened 2021-09-06
Make soak tests work again now that we subscribe by genesis hash

jsdw:
A small fix to the soak tests to make them work again using genesis hash to subscribe.


maciejhirsz (approved):
LGTM

dvdplm (approved):

PR #402, opened 2021-09-17
adding dual deployment gitlab-ci

arshamparity:
The Gitlab-CI has three stages:
- **build**
  - builds, tags and pushes Docker images to the Docker repo.
  - When a new tag is pushed (like Github releases), CI tags docker images with the commit tag.
  - When a new commit happens on the Master branch (without any tags!), CI tags docker images with `${CI_COMMIT_SHORT_SHA}-beta`
- **deploy-staging**
  - automatically deploys the images to the staging environment for QA tests.
  - Runs when a new tag is introduced or a commit has happened to the Master branch
- **deploy-production**
  - Runs only on tags
  - The deployment is manual and someone should hit the deploy button to deploy to production

P.S. The deploy-production stage is disabled atm until the production environment gets ready.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -30,9 +30,9 @@ variables:
   FORCE_DOCKER_TAG:           ""        # choose an existing docker tag to be deployed (e.g. v1.2.3)
 
   # Vault variables (Optional)
-  VAULT_SERVER_URL:               "https://vault.parity-mgmt-vault.parity.io"
-  VAULT_AUTH_PATH:                "gitlab-parity-io-jwt"
-  VAULT_AUTH_ROLE:                "cicd_gitlab_parity_${CI_PROJECT_NAME}"
+  VAULT_SERVER_URL:           "https://vault.parity-mgmt-vault.parity.io"
+  VAULT_AUTH_PATH:            "gitlab-parity-io-jwt"
+  VAULT_AUTH_ROLE:            "cicd_gitlab_parity_${CI_PROJECT_NAME}"
 
 
 default:
@@ -80,8 +80,8 @@ stages:
         --format=docker \
         --tag "$CONTAINER_REPO:$DOCKER_IMAGE_TAG" "$DOCKERFILE_DIRECTORY"
       fi
-    - echo ${Docker_Hub_Pass_Parity} |
-        buildah login --username ${Docker_Hub_User_Parity} --password-stdin docker.io
+    - echo ${DOCKER_HUB_PASS} |
+        buildah login --username ${DOCKER_HUB_USER} --password-stdin docker.io
     - |-
       echo pushing "$CONTAINER_REPO:$DOCKER_IMAGE_TAG"
       if [[ $BUILD_LATEST_IMAGE ]]; then
@@ -142,12 +142,14 @@ build-backend:
     CONTAINER_REPO:       "docker.io/parity/substrate-telemetry-backend"
     DOCKERFILE_DIRECTORY: "./backend/"
   <<:                     *dockerize
+  <<:                     *vault-secrets
 
 build-frontend:
   variables:
     CONTAINER_REPO:       "docker.io/parity/substrate-telemetry-frontend"
     DOCKERFILE_DIRECTORY: "./frontend/"
   <<:                     *dockerize
+  <<:                     *vault-secrets
 
 deploy-staging:
   variables:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,86 +1,173 @@
+# Gitlab-CI Workflow
+# stages:
+#   build:
+#     - Runs on commits on master or tags that match the pattern "v[0-9]+\.[0-9]+.*$". (e.g. 1.0, v2.1rc1)
+#   deploy-staging: 
+#     - Runs on commits on master or tags that match the pattern v1.0, v2.1rc1 (continues deployment)
+#   deploy-production:
+#     - Runs on tags that match the pattern v1.0, v2.1rc1 (manual deployment)
+
 variables:
-  BACKEND_CONTAINER_REPO:         "docker.io/parity/substrate-telemetry-backend"
-  FRONTEND_CONTAINER_REPO:        "docker.io/parity/substrate-telemetry-frontend"
-  BACKEND_IMAGE_FULL_NAME:        "${BACKEND_CONTAINER_REPO}:${CI_COMMIT_SHORT_SHA}-beta"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -1,86 +1,175 @@
...
+  # Vault variables (Optional)
+  VAULT_SERVER_URL:           "https://vault.parity-mgmt-vault.parity.io"
+  VAULT_AUTH_PATH:            "gitlab-parity-io-jwt"
+  VAULT_AUTH_ROLE:            "cicd_gitlab_parity_${CI_PROJECT_NAME}"
+
+
+default:
+  before_script:
+    - |-
+      echo defining DOKCER_IMAGE_TAG variable
```
jsdw, line 41:
```suggestion
      echo defining DOCKER_IMAGE_TAG variable
```

.gitlab-ci.yml
```diff
@@ -1,86 +1,175 @@
...
+      if [[ $FORCE_DOCKER_TAG ]]; then
+        export DOCKER_IMAGE_TAG="${FORCE_DOCKER_TAG}"
+      elif [[ $CI_COMMIT_TAG =~ ^v[0-9]+\.[0-9]+.*$ ]]; then
+        export DOCKER_IMAGE_TAG="${CI_COMMIT_TAG}"
+        #export BUILD_LATEST_IMAGE="true"
```
jsdw, line 46: Is this supposed to be commented out? There's a reference to it below, but I can't see how the variable would otherwise exist (unless it comes from the UI, and if so maybe it's worth documenting as you have done FORCE_DEPLOY etc?) :)

arshamparity: I've commented it out intentionally, until the production is not ready we shouldn't change overwrite the latest tag in the docker repository.
once everything is tested and production env is prepared we can uncomment this and publish the new docker image as the latest tag.

.gitlab-ci.yml
```diff
@@ -1,86 +1,175 @@
...
+        buildah push --format=v2s2 "$CONTAINER_REPO:$DOCKER_IMAGE_TAG"
+      fi
   rules:
-    - if: '$CI_COMMIT_BRANCH == "master"'
+    - if: '$FORCE_DOCKER_TAG'
+      when: never
+    - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+.*$/'         # i.e. v1.0, v2.1rc1
+      when: always
+    - if: '$CI_COMMIT_REF_NAME == "master" || $CI_COMMIT_REF_NAME == "main"'
       when: always
```
jsdw, line 99: do the reuls get evaluated one after the other, so if FORCE_DOCKER_TAG is true, we'll never build the images, or do they all get individually checked and _any_ true result leads to the step being run?

.gitlab-ci.yml
```diff
@@ -1,86 +1,175 @@
...
 
-.deploy-k8s:              &deploy-k8s
+.deploy:                  &deploy
   image:                  paritytech/kubetools:3.5.3
   script:
     - |-
-      helm repo add parity https://paritytech.github.io/helm-charts/
-      helm repo update
-      kubectl get cm helm-custom-values -n $KUBE_NAMESPACE -o jsonpath='{.data.values-parity\.yaml}' > values-parity.yaml
+      echo genenrating an empty custom-values.yaml file
```
jsdw, line 107:
```suggestion
      echo generating an empty custom-values.yaml file
```

.gitlab-ci.yml
arshamparity: >  so if FORCE_DOCKER_TAG is true, we'll never build the images
yes, exactly!

.gitlab-ci.yml
arshamparity: i don't see any changes here.

jsdw: The "KC" has become "CK" :)

(sometimes github sucks at updating stuff, so maybe it's not showing up properly!)

.gitlab-ci.yml
jsdw: Cool, thanks!

.gitlab-ci.yml
```diff
@@ -1,86 +1,175 @@
...
+    CONTAINER_REPO_FRONTEND: "docker.io/parity/substrate-telemetry-frontend"
+  stage:                  deploy-staging
+  <<:                     *deploy
   environment:
     name:                 parity-stg
+  rules:
+    - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+.*$/'       # i.e. v1.0, v2.1rc1
+    - if: '$CI_COMMIT_REF_NAME == "master" || $CI_COMMIT_REF_NAME == "main"'
+
+#deploy-production:
```
jsdw, line 166: I geuss this is also commented out for safety until we are ready to go live?

jsdw (approved):
I'm no expert with CI bits but this looks reasonable to me!

I'm not sure yet what the GitLab UI will look like for working with this, and so I don't yet know how easy it will be to trigger a build to live with the correct variable values and such?

.gitlab-ci.yml
jsdw: Ah yes, you said so in your PR message :)

.gitlab-ci.yml
arshamparity: good catch 👍

gabreal (approved):
very nice!

arshamparity:
> I'm no expert with CI bits but this looks reasonable to me!
> 
> I'm not sure yet what the GitLab UI will look like for working with this, and so I don't yet know how easy it will be to trigger a build to live with the correct variable values and such?

@jsdw for deploying to live you could easily set "FORCE_DEPLOY=true" in the [gitlab-ci web console](https://gitlab.parity.io/parity/substrate-telemetry/-/pipelines/new). This will make the `deploy-production` job available to be played manually by clicking the play button. However this will build docker images from the latest commit from the master branch. if you want to specify which docker tag you want to be deployed, the `FORCE_DOCKER_TAG=v1.0.21` should also be used. Then:
- the build stage would be skipped as you are using an existing docker image
- the deploy-staging job would be triggered and played automatically.
- the deploy-production stage would be triggered and could be deployed manually by pressing the play button

So both of these variables could be used explicitly depending on your needs. They could also be used together like the above example.

dvdplm:
@arshamparity ty! :)

Can we get the deployment instructions added to the README in this repo or somewhere else where it's easy to find?

arshamparity:
@dvdplm this Gitlab-CI file is going to be used as a template for our further CIs. This is the first pipeline that is using that template. I'm going to add a wiki page about it so this information could be found there.
The shorter version of this information is also available in the GitLab-CI file where these variables are defined. If that's not clear, that should also be rewritten. (will probably do after the migration)

PR #404, opened 2021-09-20
Deduplicate multiplexed nodes by network_id

### LINKED ISSUE #403: Khala Network peers bringing huge amount of dummy nodes in Telemetry
We have observed a huge amount of nodes showing on Kusama Telemetry page. Till the end of yesterday, we have seen 900k nodes. Most of the newly added nodes are at block #0, and each node shows for 10-20 times with the same name.

We suspect most of them come from Khala clients. Before we launched our secure enclave mining, in my impression the Kusama Telemetry board looks normal (no confirmation). However after our launch, it quickly goes beyond 100k, and approached 900k nodes yesterday. Some nodes have the nodes indicating they are Khala miners.

Khala mining requires each miner to sync a Khala full node, which comes with a full Kusama node as well. So theoretically, we should bring the amount of Kusama nodes up to the number of our miners. However the Telemetry shows duplicated miners and out-of-date nodes, which is problematic.

This is our Khala chain-spec where the telemetry endpoint was defined. It should load the unmodified Kusama chain-spec in a parachain client as well:

<https://github.com/Phala-Network/khala-parachain/blob/main/node/res/khala.json>

maciejhirsz:
Should fix #403.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -286,6 +286,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 
 [[package]]
+name = "arrayvec"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


dvdplm (approved):
Let's try this and see what happens.

jsdw (approved):

PR #406, opened 2021-09-20
Tidy up stale connections.

jsdw:
This PR adds logic to tidy stale node connections. It will:

- Remove nodes when telemetry corresponding to the node message ID isn't seen for a while.
- Disconnect telemetry connections that haven't sent useful telemetry data in a while.

To accomodate this, I had to spawn our receiving of socket data onto a separate task, because it's not cancel-safe and so can't be interrupted by the periodic check that I wanted to do. (The alternative is using BiLocks and such and spawning out the periodic interval instead, but that was more cumbersome!)

### DIFF
--- a/backend/telemetry_shard/src/aggregator.rs
+++ b/backend/telemetry_shard/src/aggregator.rs
@@ -73,6 +73,10 @@ pub enum FromWebsocket {
         message_id: node_message::NodeMessageId,
         payload: node_message::Payload,
     },
+    /// remove a node with the given message ID
+    Remove {
+        message_id: node_message::NodeMessageId,
+    },
     /// Make a note when the node disconnects.
     Disconnected,
 }
@@ -246,7 +250,28 @@ impl Aggregator {
                     let _ = tx_to_telemetry_core
                         .send_async(FromShardAggregator::UpdateNode { local_id, payload })
                         .await;
-                }
+                },
+                ToAggregator::FromWebsocket(
+                    conn_id,
+                    FromWebsocket::Remove {
+                        message_id
+                    }
+                ) => {
+                    // Get the local ID, ignoring the message if none match:
+                    let local_id = match to_local_id.get_id(&(conn_id, message_id)) {
+                        Some(id) => id,
+                        None => continue,
+                    };
+
+                    // Remove references to this single node:
+                    muted.remove(&local_id);
+                    to_local_id.remove_by_id(local_id);
+
+                    // Tell the core to remove references to it, too:
+                    let _ = tx_to_telemetry_core
+                        .send_async(FromShardAggregator::RemoveNode { local_id })
+                        .await;
+                },
                 ToAggregator::FromWebsocket(disconnected_conn_id, FromWebsocket::Disconnected) => {
                     // Find all of the local IDs corresponding to the disconnected connection ID and
                     // remove them, telling Telemetry Core about them too. This could be more efficient,
--- a/backend/telemetry_shard/src/aggregator.rs
+++ b/backend/telemetry_shard/src/aggregator.rs
@@ -250,13 +250,8 @@ impl Aggregator {
                     let _ = tx_to_telemetry_core
                         .send_async(FromShardAggregator::UpdateNode { local_id, payload })
                         .await;
-                },
-                ToAggregator::FromWebsocket(
-                    conn_id,
-                    FromWebsocket::Remove {
-                        message_id
-                    }
-                ) => {
+                }
+                ToAggregator::FromWebsocket(conn_id, FromWebsocket::Remove { message_id }) => {
                     // Get the local ID, ignoring the message if none match:
                     let local_id = match to_local_id.get_id(&(conn_id, message_id)) {
                         Some(id) => id,
@@ -271,7 +266,7 @@ impl Aggregator {
                     let _ = tx_to_telemetry_core
                         .send_async(FromShardAggregator::RemoveNode { local_id })
                         .await;
-                },
+                }
                 ToAggregator::FromWebsocket(disconnected_conn_id, FromWebsocket::Disconnected) => {
                     // Find all of the local IDs corresponding to the disconnected connection ID and
                     // remove them, telling Telemetry Core about them too. This could be more efficient,
--- a/backend/telemetry_shard/src/main.rs
+++ b/backend/telemetry_shard/src/main.rs
@@ -34,7 +34,7 @@ use common::http_utils;
 use common::node_message;
 use common::node_message::NodeMessageId;
 use common::rolling_total::RollingTotalBuilder;
-use futures::SinkExt;
+use futures::{ SinkExt, StreamExt };
 use http::Uri;
 use hyper::{Method, Response};
 use simple_logger::SimpleLogger;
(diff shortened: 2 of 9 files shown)


dvdplm:
One concern I have is about what happens in a crisis for a network, when nodes behave erratically, go off on forks, struggle to produce blocks and generally are just sloooooow. In such times the community tends to flock to telemetry to get a sense of where things are going. With this in place, do we risk evicting stuck nodes from telemetry as well as stale/spurious ones?

backend/telemetry_shard/src/aggregator.rs
```diff
@@ -247,6 +251,22 @@ impl Aggregator {
                         .send_async(FromShardAggregator::UpdateNode { local_id, payload })
                         .await;
                 }
+                ToAggregator::FromWebsocket(conn_id, FromWebsocket::Remove { message_id }) => {
+                    // Get the local ID, ignoring the message if none match:
+                    let local_id = match to_local_id.get_id(&(conn_id, message_id)) {
+                        Some(id) => id,
+                        None => continue,
+                    };
+
```
dvdplm, line 260: Do you think we should collect stats on the nr of ignored messages?

backend/telemetry_shard/src/aggregator.rs
```diff
@@ -246,7 +250,28 @@ impl Aggregator {
...
+                    // Remove references to this single node:
+                    muted.remove(&local_id);
+                    to_local_id.remove_by_id(local_id);
+
+                    // Tell the core to remove references to it, too:
+                    let _ = tx_to_telemetry_core
+                        .send_async(FromShardAggregator::RemoveNode { local_id })
+                        .await;
```
dvdplm, line 268: Might be worth to put into its own utility method so we can rest assured none of these steps are ever skipped (I think it's repeated below as well?).

backend/telemetry_shard/src/main.rs
```diff
@@ -283,21 +289,28 @@ where
...
+
+                    // Node with network ID already exist? Remove it to be replaced with this one.
+                    if let Some(&old_message_id) = network_id_to_msg_id.get(&network_id) {
+                        network_id_to_msg_id.remove(&network_id);
+                        allowed_message_ids.remove(&old_message_id);
+                        let _ = tx_to_aggregator.send(FromWebsocket::Remove { message_id: old_message_id }).await;
+                    }
+                    // Too many nodes seen on this connection? Ignore this one.
+                    else if network_id_to_msg_id.len() >= max_nodes_per_connection {
+                        continue;
```
dvdplm, line 348: Should we collect stats on this so we know if/when things go haywire?

backend/telemetry_shard/src/main.rs
```diff
@@ -91,6 +96,12 @@ struct Opts {
     /// on the machine. If no value is given, use an internal default that we have deemed sane.
     #[structopt(long)]
     worker_threads: Option<usize>,
+    /// Roughly how long to wait in seconds for new telemetry data to arrive from a node. If
+    /// telemetry for a node does not arrive in this time frame, we remove the corresponding node
+    /// state, and if no messages are received on the connection at all in this time, it will be
+    /// dropped.
+    #[structopt(long, default_value = "30")]
```
dvdplm, line 103: I think we might need a longer time here to account for nodes that are "legitimately silent" due to averse network conditions. Tbh I'm actually not sure if stuck nodes send telemetry messages or not, i.e. if it's possible/plausible that a node stays silent because it's having internal issues of some sort?

backend/telemetry_shard/src/main.rs
```diff
@@ -200,10 +213,15 @@ async fn handle_node_websocket_connection<S>(
     max_nodes_per_connection: usize,
     bytes_per_second: ByteSize,
     block_list: BlockedAddrs,
+    stale_node_timeout: Duration,
 ) -> (S, http_utils::WsSender)
 where
     S: futures::Sink<FromWebsocket, Error = anyhow::Error> + Unpin + Send + 'static,
 {
+    // Keep track of the message Ids that have been "granted access". We allow a maximum of
+    // `max_nodes_per_connection` before ignoreing others.
```
dvdplm, line 222:
```suggestion
    // `max_nodes_per_connection` before ignoring others.
```

backend/telemetry_shard/src/main.rs
```diff
@@ -212,46 +230,84 @@ where
...
-                if let Err(soketto::connection::Error::Closed) = msg_info {
-                    break;
+            // We periodically check for stale message IDs and remove nodes associated with
+            // them, to prevent a buildup. We boot the whole connection if no interpretable
+            // messages have been sent at all in the time period.
+            _ = stale_interval.tick() => {
+                let stale_ids: Vec<NodeMessageId> = allowed_message_ids.iter()
+                    .filter(|(_, last_seen)| last_seen.elapsed() > stale_node_timeout)
+                    .map(|(&id, _)| id)
+                    .collect();
```
dvdplm, line 292: Do we really need to `collect` the ids or can we do all the work while we iterate through them?

dvdplm (approved):
Code lgtm, left some nits.

I do feel a wee bit uncomfortable about not fully understanding what is causing the problem we're solving here. The code is fine, and it could very well be that we need to do this, but I still think we need to dig into what is causing nodes to spam us with fresh message IDs. :|

backend/telemetry_shard/src/main.rs
jsdw: Yeahh, we need to collect them at the mo because we can't remove them from the `allowed_message_ids` while iterating. There is a lovely function I wanted to use called `drain_filter` that would have let me iterate over IDs being removed and do something with them, but it's not stable yet!

backend/telemetry_shard/src/main.rs
jsdw: I added a log message here at least; collecting stats on it would be a little trickier at the mo!

backend/telemetry_shard/src/aggregator.rs
jsdw: I agree (and I tweaked the code so those 2 sections look identical); I'd need to restructure things a little to make that be the case though (or do something horrible like a macro), because those maps are just local variables, so I'm leaning towards not doing that right now.

backend/telemetry_shard/src/aggregator.rs
jsdw: We do collect stats on the number of messages we ignore because there are too many to deal with at the mo. This code just silently ignores messages from Ids that don't exist incase nodes are removed while messages that work on them are still queued for some reason (I'm not sure if I'd ever actually expect to encounter the "None" case though)

niklasad1 (approved):
I agree with @dvdplm's comments this is a bit opinionated, the code looks good otherwise.

Are we really sure that closing the connection on the client side works properly?

backend/telemetry_shard/src/main.rs
```diff
@@ -91,6 +96,12 @@ struct Opts {
     /// on the machine. If no value is given, use an internal default that we have deemed sane.
     #[structopt(long)]
     worker_threads: Option<usize>,
+    /// Roughly how long to wait in seconds for new telemetry data to arrive from a node. If
+    /// telemetry for a node does not arrive in this time frame, we remove the corresponding node
+    /// state, and if no messages are received on the connection at all in this time, it will be
+    /// dropped.
+    #[structopt(long, default_value = "60")]
+    stale_node_timeout: u64,
```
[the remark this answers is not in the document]
jsdw: We aren't currently sending back any sort of close reason (I'm not sure whether that's possible/should be added now, but when I started this stuff I think it wasn't a thing yet?).

I'm not aware of a message that substrate would send to keep the connection open either! This currently relies on the assumption that if we don't see any new telemetry for a while, something has probably gone wrong on their side. I'm not sure how true that is, or whether it's perfectly reasonable for a node to not send any telemetry for over a minute!

jsdw:
One potential issue with this PR is that if the connection stops sending telemetry about a node for over a minute (current default), we'll clean up our state (fine), and then if the connection starts sending messages about the node again we won't re-add the node, because we would need a new system.connected message to do so.

(This could be addressed by holding on to system.connected messages from a connection, but that increases complexity more.)

If we wanted, we could tweak this PR to allow "stale" nodes to exist, but still kill the connection entirely if no useful telemetry is sent at all within the time. This may not address an issue with lots of nodes being advertised on a connection (if that's even the problem), but at least inactive connections would be shutdown, and substrate nodes will try reconnecting by default anyway, so a healthy node can recover from that without a restart.

PR #414, opened 2021-09-29
Add internal deployment steps to the README

jsdw:
Just so that we know how to deploy this thing in the future :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,3 +171,41 @@ The building process is standard. You just need to notice that the `Dockerfile`s
 DOCKER_USER=$USER ./scripts/build-docker-frontend.sh
 DOCKER_USER=$USER ./scripts/build-docker-backend.sh
 ```
+
+## Deployment
+
+This section covers the internal deployment of Substrate Telemetry to our staging and live environments.
+
+### Deployment to staging
+
+Every time new code is merged to `master`, a new version of telemetry will be built and deployed to our staging environment. Roughly what will happen is:
+
+- An image tag will be generated that looks like `$CI_COMMIT_SHORT_SHA-beta`, for example `224b1fae-beta`.
+- Docker images for the frontend and backend will be pushed to the docker repo (see https://hub.docker.com/r/parity/substrate-telemetry-backend/tags?page=1&ordering=last_updated and https://hub.docker.com/r/parity/substrate-telemetry-frontend/tags?page=1&ordering=last_updated).
+- A deployment to the staging environment will be performed using these images. Go to https://gitlab.parity.io/parity/substrate-telemetry/-/pipelines to inspect the progress of such deployments.
+
+### Deployment to live
+
+Once we're happy with things in staging, we can do a deployment to live as follows.
+
+1. Ensure that the PRs you'd like to deploy are merged to master.
+2. Tag the commit that you'd like to deploy:
+   - Run `git tag | grep -e '^v[0-9]\+\.[0-9]\+.*$'` to see which tags currently exist that meet the required format (tags not matching this format will not trigger the deployment process).
+   - Use eg `git tag v2.0-foo` to add a tag with a version number that has had the major version incremented from those tags found above. The suffix `-foo` can be omitted entirely or can contain some information that you feel would be useful.
+   - Push the tag to master with eg `git push origin v2.0-foo`. If this fails, you may need permissions on the repository elevating.
+3. In Gitlab, you can view the progress of the deployment at https://gitlab.parity.io/parity/substrate-telemetry/-/pipelines.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -171,3 +171,41 @@ The building process is standard. You just need to notice that the `Dockerfile`s
...
+5. Visit the CI/CD pipelines page again (URl above) and click the "play" button on the "Deploy-production" stage to perform the deployment to live.
+6. Confirm that things are working once the deployment has finished by visiting https://telemetry.polkadot.io/.
+
+### Rolling back to a previous deployment
+
+If something goes wrong running the above, we can roll back the deployment to live as follows.
+
+1. Decide what image tag you'd like to roll back to. Go to https://hub.docker.com/r/parity/substrate-telemetry-backend/tags?page=1&ordering=last_updated and have a look at the available tags (eg `v2.0-224b1fa`) to select one you'd like. You can cross reference this with the tags available using `git tag` in the repository to help see which tags correspond to which code changes.
+2. Navigate to https://gitlab.parity.io/parity/substrate-telemetry/-/pipelines/new.
+3. Add a variable called `FORCE_DEPLOY` with the value `true`.
```
arshamparity, line 207: I think it worths specifying that `FORCE_DEPLOY` tag could also be used alone if someone wants to activate the production pipeline with the latest commit on that master.

arshamparity (commented):
The readme looks great!
I might use it for documenting our GitLab ci file as well.

README.md
jsdw: Ah, gotcha! So if I understand right, and looking at the CI file, then if only `FORCE_DEPLOY` is used, `DOCKER_IMAGE_TAG` will end up equal to `${CI_COMMIT_SHORT_SHA}-beta` (will `CI_COMMIT_SHORT_SHA` be equal to the latest commit hash if we run this manually?). So, we'd end up deploying the latest commit as you say, which uses the images that would be currently running in staging.

That's good to know. I almost wonder whether that should be our default process as it's simpler than pushing a tag to master..

jsdw: I had a chat with the team and actually we prefer using FORCE_DEPLOY=true as the way to deploy to live anyway (rather than worrying about version numbers), so I'll update the readme to talk about that more and de-emphasise the version number bits.

README.md
```diff
@@ -171,3 +171,49 @@ The building process is standard. You just need to notice that the `Dockerfile`s
...
+
+## Deployment
+
+This section covers the internal deployment of Substrate Telemetry to our staging and live environments.
+
+### Deployment to staging
+
+Every time new code is merged to `master`, a new version of telemetry will be automatically built and deployed to our staging environment, so there is nothing that you need to do. Roughly what will happen is:
+
+- An image tag will be generated that looks like `$CI_COMMIT_SHORT_SHA-beta`, for example `224b1fae-beta`.
```
dvdplm, line 183: @arshamparity What is the idea for the `-beta` postfix? Is that something you want to keep long term or was it more of a temporary thing. To us the staging env is useful to manually test new code, but as we do not plan on doing proper releases for `substrate-telemetry` for the time being, there will also not be "beta versions" released. So is the `-beta` postfix still useful?

README.md
```diff
@@ -171,3 +171,49 @@ The building process is standard. You just need to notice that the `Dockerfile`s
...
+- Docker images for the frontend and backend will be pushed to the docker repo (see https://hub.docker.com/r/parity/substrate-telemetry-backend/tags?page=1&ordering=last_updated and https://hub.docker.com/r/parity/substrate-telemetry-frontend/tags?page=1&ordering=last_updated).
+- A deployment to the staging environment will be performed using these images. Go to https://gitlab.parity.io/parity/substrate-telemetry/-/pipelines to inspect the progress of such deployments.
+
+### Deployment to live
```
dvdplm, line 187: I suspect it's more accurate to s/live/production/ but I could be wrong about which is the more common term.

dvdplm (approved):

README.md
arshamparity: the idea was to build the images first and test them on staging environment with `-beta` postfix, once we are sure the image is working fine, a new tag should be pushed to GitHub, then that version would be used to build the same docker image with the tag provided. we don't need to necessarily do a release, only a new tag would be sufficient to activate the production pipeline. 
With this logic, public users would be informed about `beta` images and they won't use them for their production as they might be broken, but when a new version without `-beta` is available, they can trust that version and deploy it on their production envs.

README.md
arshamparity: afaict, we used production for namings. So I think it might be good to use the same string for the documentation.

arshamparity (approved):
looks great to me!

README.md
jsdw: Done!

PR #415, opened 2021-09-29
Jsdw no ci versions

jsdw:
As a team we decided that it's easier not to think about version numbering when deploying.

I updated the README to reflect that in #414, but we could go one step further and simplify the CI file (and further simplify the README) if we have no interest in version numbering.

This makes releasing to live as simple as merging a PR to master and then, if we're happy with it in staging, going and pressing a "play" button in the CI to deploy it to live. We maintain the ability to roll back to previous images.


jsdw:
We've decided to stick with version numbering now for CI consistency with other projects, so I'll close this.

PR #416, opened 2021-09-29
Expose total messages sent to aggregator, too

jsdw:
This will allow us to graph the dropped messages per interval as well as seeing how many messages are coming in and being dropped in total.

### DIFF
--- a/backend/telemetry_core/src/aggregator/inner_loop.rs
+++ b/backend/telemetry_core/src/aggregator/inner_loop.rs
@@ -238,7 +238,7 @@ impl InnerLoop {
                         tx,
                         metered_rx.len(),
                         dropped_messages2.load(Ordering::Relaxed),
-                        total_messages2.load(Ordering::Relaxed)
+                        total_messages2.load(Ordering::Relaxed),
                     ),
                 }
             }
--- a/backend/telemetry_core/src/aggregator/inner_loop.rs
+++ b/backend/telemetry_core/src/aggregator/inner_loop.rs
@@ -119,8 +119,10 @@ pub struct Metrics {
     /// How many messages are currently queued up in internal channels
     /// waiting to be sent out to feeds.
     pub total_messages_to_feeds: usize,
-    /// How many messages are queued waiting to be handled by this aggregator.
-    pub total_messages_to_aggregator: usize,
+    /// How many messages are currently queued waiting to be handled by this aggregator.
+    pub current_messages_to_aggregator: usize,
+    /// The total number of messages sent to the aggregator.
+    pub total_messages_to_aggregator: u64,
     /// How many (non-critical) messages have been dropped by the aggregator because it was overwhelmed.
     pub dropped_messages_to_aggregator: u64,
     /// How many nodes are currently known to this aggregator.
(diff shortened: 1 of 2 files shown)


backend/telemetry_core/src/main.rs
```diff
@@ -537,6 +537,10 @@ async fn return_prometheus_metrics(aggregator: AggregatorSet) -> Response<hyper:
             "telemetry_total_messages_to_feeds{{aggregator=\"{}\"}} {} {}\n",
             idx, m.total_messages_to_feeds, m.timestamp_unix_ms
         ));
+        s.push_str(&format!(
+            "telemetry_current_messages_to_aggregator{{aggregator=\"{}\"}} {} {}\n\n",
+            idx, m.current_messages_to_aggregator, m.timestamp_unix_ms
+        ));
```
maciejhirsz, line 543: You can change all of these to:

```suggestion
        write!(
            &mut s,
            "telemetry_current_messages_to_aggregator{{aggregator=\"{}\"}} {} {}\n\n",
            idx, m.current_messages_to_aggregator, m.timestamp_unix_ms
        );
```

backend/telemetry_core/src/aggregator/inner_loop.rs
```diff
@@ -211,13 +213,15 @@ impl InnerLoop {
         let max_queue_len = self.max_queue_len;
         let (metered_tx, metered_rx) = flume::unbounded();
 
-        // Keep count of the number of messages we drop for the sake of metric reporting
+        // Keep count of the number of dropped/total messages for the sake of metric reporting
         let dropped_messages = Arc::new(AtomicU64::new(0));
+        let total_messages = Arc::new(AtomicU64::new(0));
```
maciejhirsz, line 218: Could put it on the same `Arc` with dropped_messages.

maciejhirsz (approved):
Minor grumbles but I wouldn't sweat it.

backend/telemetry_core/src/aggregator/inner_loop.rs
jsdw: That's true; I'll do that actually!

backend/telemetry_core/src/main.rs
jsdw: Ooh yeah, that is better, thanks!

backend/telemetry_core/src/main.rs
```diff
@@ -537,6 +537,10 @@ async fn return_prometheus_metrics(aggregator: AggregatorSet) -> Response<hyper:
             "telemetry_total_messages_to_feeds{{aggregator=\"{}\"}} {} {}\n",
             idx, m.total_messages_to_feeds, m.timestamp_unix_ms
         ));
+        s.push_str(&format!(
+            "telemetry_current_messages_to_aggregator{{aggregator=\"{}\"}} {} {}\n\n",
```
dvdplm, line 541: Do you think it's worth ghetto-namespacing the label with "core" somewhere, so it's crystal clear what this refers to? Maybe `telemetry_core_current_messages_to_aggregator` or `telemetry_current_messages_to_core_aggregator`?

backend/telemetry_core/src/main.rs
```diff
@@ -537,6 +537,10 @@ async fn return_prometheus_metrics(aggregator: AggregatorSet) -> Response<hyper:
...
+            idx, m.current_messages_to_aggregator, m.timestamp_unix_ms
+        ));
         s.push_str(&format!(
             "telemetry_total_messages_to_aggregator{{aggregator=\"{}\"}} {} {}\n\n",
```
dvdplm, line 545: Same here, perhaps good to add "core" somewhere?

dvdplm (approved):
lgtm

backend/telemetry_core/src/main.rs
jsdw: That's a good idea; I guess I haven't done that yet because all of the telemetry is currently from/about the core

backend/telemetry_core/src/aggregator/inner_loop.rs
jsdw: I had a go at this but decided not to do it in the end just because it made the code messier :)

backend/telemetry_core/src/main.rs
jsdw: I namespaced everything with "telemetry_core" in the end so it'll make more sense if/when we have shard metrics, too! I'll update the graphs once things are deployed.

backend/telemetry_core/src/main.rs
jsdw: Done!

Issue #417, opened 2021-09-29
Telemetry can get stuck into a state of being constantly overwhelmed when core is underpowered


jsdw:
Telemetry ran into an issue on 28th Sept starting at approx 11:30 UTC. (internal metrics are at https://grafana.parity-mgmt.parity.io/d/vnlkQiInk/substrate-telemetry-metrics?orgId=1&var-datasource=prometheus.parity-prod&from=1632823200000&to=1632841200000 and resource usage at https://grafana.parity-mgmt.parity.io/d/85a562078cdf77779eaa1add43ccec1e/kubernetes-compute-resources-namespace-pods?orgId=1&var-datasource=prometheus.parity-prod&var-cluster=&var-namespace=substrate-telemetry&var-interval=4h&from=1632823200000&to=1632841200000).

Our theory is:

1. For some reason (here, possibly a new chain being launched) the Telemetry core had a sudden backlog of messages to deal with.
2. Pings from UIs weren't responded to quickly enough due to this backlog, so UIs would try to reconnect and resubscribe every ~30 secs (I'm not sure on what the reconnect logic is, but pings are every 30s I think)
3. The UIs all reconnecting at a similar time led to a bunch of subscription requests (for Kusama too owing to an auto subscribe thing that's now fixed) which overburdened the telemetry core again
4. Repeat from step 2.

(This is all a bit thoretical but I think it would explain why both the CPU usage and memory usage increased (lots of subscribing work to do), and the subscribed feeds dropped off (they were reconnecting after not receiving ping responses in time)).

To handle subscribes efficiently (a lot of data needs serializing and sending out each subscribe) the telemetry core process can fully utilise however many cores are available to it, but I'm guessing it doesn't have that many available in the current setup, so this sort of bursty stuff is hard to handle.

@dvdplm  pointed out that the reconnect logic in the UI could likely be improved with eg an exponential backoff and some randomisation to help spread the re-subscribe workload out.

"Pings" currently make a roundtrip through the telemetry core, which can mean they can be stuck behind other messages that need handling. This makes them useful to test that the system is responding well, but of course can lead to the above if it's being slow, and make the problem worse.

I think looking at the reconnect logic first will be potentially the quickest win and most helpful change, and #416 helps a little with the metrics around dropped messages, to get a feel for the absolute impact of this.

jsdw:
I haven't seen this crop up again since #418 merged, so will close this for now!

PR #418, opened 2021-09-30
Timeout based on last msg received and not ping timeout

jsdw:
To help ease #417, this PR makes the UI timeout not be based on how long it takes to get a ping response, but on how long it takes to get back _any_ message from the backend (if messages are still being received, we assume that the connection is still active; it doesn't necesasrily have to be a ping response). I also allow 60 seconds from the last message seen, though we might feel that that is a bit high.

We still ping similarly to before to before in order to keep a connection alive if no other messages are being sent (eg UI not subscribed to anything), though since ping no longer determines success or fail, we only send the next ping out when the previous one returns.

Questions:
- Do we prefer to reconnect based on the ping timeout, to test that the entire round trip of the connection is OK and not just that we are still getting messages back OK (It feels like checking for msgs back is good enough to me)?
- Would we prefer an exponential reconnect? (This was my original plan, but it wasn't obvious in the current code how best to do it offhand; currently the disconnect-on-ping-timeout is sortof a by product of just trying to send another ping before the last one has come back, and we try to reconnect immediately on disconnect).

### DIFF
--- a/frontend/src/Connection.ts
+++ b/frontend/src/Connection.ts
@@ -33,8 +33,9 @@ import {
   StateCacheColumn,
 } from './components/List';
 
-const TIMEOUT_BASE = (1000 * 5) as Types.Milliseconds; // 5 seconds
-const TIMEOUT_MAX = (1000 * 60 * 5) as Types.Milliseconds; // 5 minutes
+const CONNECTION_TIMEOUT_BASE = (1000 * 5) as Types.Milliseconds; // 5 seconds
+const CONNECTION_TIMEOUT_MAX = (1000 * 60 * 5) as Types.Milliseconds; // 5 minutes
+const MESSAGE_TIMEOUT = (1000 * 60) as Types.Milliseconds; // 60 seconds
 
 declare global {
   interface Window {
@@ -74,12 +75,15 @@ export class Connection {
 
   private static async socket(): Promise<WebSocket> {
     let socket = await Connection.trySocket();
-    let timeout = TIMEOUT_BASE;
+    let timeout = CONNECTION_TIMEOUT_BASE;
 
     while (!socket) {
       await sleep(timeout);
 
-      timeout = Math.min(timeout * 2, TIMEOUT_MAX) as Types.Milliseconds;
+      timeout = Math.min(
+        timeout * 2,
+        CONNECTION_TIMEOUT_MAX
+      ) as Types.Milliseconds;
       socket = await Connection.trySocket();
     }
 
@@ -112,6 +116,8 @@ export class Connection {
     });
   }
 
+  // timer which will force a reconnection if no message is seen for a while
+  private messageTimeout: Maybe<ResettableTimeout> = null;
   // id sent to the backend used to pair responses
   private pingId = 0;
   // timeout handler for ping messages
@@ -170,7 +176,8 @@ export class Connection {
     this.socket.send(`no-more-finality:${chain}`);
   }
 
-  public handleMessages = (messages: FeedMessage.Message[]) => {
+  private handleMessages = (messages: FeedMessage.Message[]) => {
+    this.messageTimeout?.reset();
     const { nodes, chains, sortBy, selectedColumns } = this.appState;
     const nodesStateRef = nodes.ref;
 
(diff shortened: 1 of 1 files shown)


frontend/src/Connection.ts
```diff
@@ -33,8 +33,9 @@ import {
   StateCacheColumn,
 } from './components/List';
 
-const TIMEOUT_BASE = (1000 * 5) as Types.Milliseconds; // 5 seconds
-const TIMEOUT_MAX = (1000 * 60 * 5) as Types.Milliseconds; // 5 minutes
+const CONNECTION_TIMEOUT_BASE = (1000 * 5) as Types.Milliseconds; // 5 seconds
+const CONNECTION_TIMEOUT_MAX = (1000 * 60 * 5) as Types.Milliseconds; // 5 minutes
```
jsdw, line 37: Just prefixing with `CONNECTION` for clarity

frontend/src/Connection.ts
```diff
@@ -33,8 +33,9 @@ import {
...
+const MESSAGE_TIMEOUT = (1000 * 60) as Types.Milliseconds; // 60 seconds
```
jsdw, line 38: If a message isn't seen in this time, we try to reconnect

frontend/src/Connection.ts
```diff
@@ -170,7 +176,8 @@ export class Connection {
     this.socket.send(`no-more-finality:${chain}`);
   }
 
-  public handleMessages = (messages: FeedMessage.Message[]) => {
+  private handleMessages = (messages: FeedMessage.Message[]) => {
+    this.messageTimeout?.reset();
```
jsdw, line 180: Reset the timeout timer each time a message is seen (and make this fn private because nothing external seemed to need access to it)

frontend/src/Connection.ts
```diff
@@ -565,3 +588,26 @@ export class Connection {
...
+
+/**
+ * Fire a function if the timer runs out. You can reset it, or
+ * cancel it to prevent the function from being fired.
+ *
+ * @param onExpired
+ * @param timeoutMs
+ * @returns
+ */
+function resettableTimeout(onExpired: () => void, timeoutMs: number) {
```
jsdw, line 600: equivalent to `setTimeout`, but it can be reset as well as cancelled

frontend/src/Connection.ts
maciejhirsz: I just went "wait, this will error if timeout is `None`" before I realized this is TypeScript.

frontend/src/Connection.ts
maciejhirsz:
```suggestion
function resettableTimeout(onExpired: () => void, timeoutMs: Types.Milliseconds) {
```

Not sure how annoying that is with the quadratic expansion function, `number` is fine if it is.

frontend/src/Connection.ts
```diff
@@ -565,3 +588,26 @@ export class Connection {
...
+  let timer = setTimeout(onExpired, timeoutMs);
+
+  return {
+    reset: () => {
+      clearTimeout(timer);
+      timer = setTimeout(onExpired, timeoutMs);
+    },
+    cancel: () => {
+      clearTimeout(timer);
+    },
```
maciejhirsz, line 610: No `this` so can just do regular method shorthand.

```suggestion
    reset() {
      clearTimeout(timer);
      timer = setTimeout(onExpired, timeoutMs);
    },
    cancel() {
      clearTimeout(timer);
    },
```

maciejhirsz (approved):
Minor grumbles

frontend/src/Connection.ts
jsdw: Ah true, I always forget about this shorthand!

(I also forget a bunch of the syntax things around classes, since I do my best to avoid them at all costs :D

frontend/src/Connection.ts
```diff
@@ -565,3 +588,29 @@ export class Connection {
...
+    reset() {
+      clearTimeout(timer);
+      timer = setTimeout(onExpired, timeoutMs);
+    },
+    cancel() {
+      clearTimeout(timer);
+    },
+  };
+}
+type ResettableTimeout = ReturnType<typeof resettableTimeout>;
```
TarikGul, line 616: Not perfectly clear on the format of the frontend, but is this the proper place to store the type?

TarikGul (approved):
lgtm, just had one question.

frontend/src/Connection.ts
jsdw: I think the types that are shoved in a separate file are mainly just the types of "incoming" data from the backend that we want to know about (or general helper types), so it feels like a good enough place to me offhand :)

PR #420, opened 2021-10-01
Saturating subtract to avoid possible underflow

jsdw:
I left a telemetry instance running locally for a while and managed to run into:

```
thread 'telemetry_core_worker' panicked at 'attempt to subtract with overflow', telemetry_core/src/state/chain.rs:262:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Ie, when doing `now - timestamp`. I'm not exactly sure how this happened offhand, but as a quick fix to prevent against future issues I swapped to using `saturating_sub`.

### DIFF
--- a/backend/telemetry_core/src/state/chain.rs
+++ b/backend/telemetry_core/src/state/chain.rs
@@ -259,7 +259,7 @@ impl Chain {
                     self.best.hash,
                 );
                 if let Some(timestamp) = self.timestamp {
-                    self.block_times.push(now - timestamp);
+                    self.block_times.push(now.saturating_sub(timestamp));
                     self.average_block_time = Some(self.block_times.average());
                 }
                 self.timestamp = Some(now);
@@ -271,7 +271,7 @@ impl Chain {
                 propagation_time = Some(0);
             } else if block.height == self.best.height {
                 if let Some(timestamp) = self.timestamp {
-                    propagation_time = Some(now - timestamp);
+                    propagation_time = Some(now.saturating_sub(timestamp));
                 }
             }
 


maciejhirsz (approved):

maciejhirsz:
[Leap second](https://en.wikipedia.org/wiki/Leap_second) maybe?

jsdw:
Ooh, interesting! Surrounding logs to get an idea of roughly when:

```
2021-09-30 13:14:00,854 INFO  [telemetry_core] Opening /feed connection from 127.0.0.1:59762
2021-09-30 16:47:28,362 INFO  [telemetry_core] Closing /feed connection from 127.0.0.1:59762
thread 'telemetry_core_worker' panicked at 'attempt to subtract with overflow', telemetry_core/src/state/chain.rs:262:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2021-09-30 18:59:46,686 INFO  [telemetry_core] Closing /shard_submit connection from 127.0.0.1:59604
2021-09-30 18:59:46,686 ERROR [telemetry_core::aggregator::inner_loop] Cannot send message into aggregator: sending on a closed channel
```

But using system time we def need to assume it won't always increase.. now to go and see about that leap second..

jsdw:
(It doesn't sound like any leap seconds happened.. my best guess now is that my system clock corrected itself at some point and that knocked it back a bit!)

maciejhirsz:
> (It doesn't sound like any leap seconds happened.. my best guess now is that my system clock corrected itself at some point and that knocked it back a bit!)

That's more likely but also more boring.

TarikGul (approved):
lgtm

PR #424, opened 2021-10-12
First/Third party by genesis hash, not label. Make limit configurable and default to 1k

jsdw:
Use genesis hash, not chain label, to determine how many nodes are allowed to be connected to a chain.
Set the limit to 1000, not 500, and make it configurable via the CLI.

### DIFF
--- a/backend/telemetry_core/src/state/chain.rs
+++ b/backend/telemetry_core/src/state/chain.rs
@@ -20,6 +20,7 @@ use common::node_types::{BlockHash, BlockNumber};
 use common::{id_type, time, DenseMap, MostSeen, NumStats};
 use once_cell::sync::Lazy;
 use std::collections::HashSet;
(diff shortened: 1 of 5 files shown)


backend/telemetry_core/src/state/chain.rs
```diff
@@ -53,6 +54,8 @@ pub struct Chain {
     timestamp: Option<Timestamp>,
     /// Genesis hash of this chain
     genesis_hash: BlockHash,
+    /// Maximum number of nodes allowed in this chain
```
dvdplm, line 57:
```suggestion
    /// Maximum number of nodes allowed to connect from this chain
```

"in" sounds a bit like it's a cap on the number of allowed nodes participating in a network?

jsdw: That's fair :)

PR #426, opened 2021-10-13
We settled on using tags to deploy to live, so reflect that in the docs

jsdw:
The docs merged previously ended up containing the wrong steps for deploying to live (they went through a bunch of iterations, and it looks like they weren't updated properly to meet the final decision; whoops). This small update fixes that.


ArshamTeymouri (approved):
looks good to me.
These also have been documented here: https://gitlab.parity.io/parity/gitlab-ci-template

dvdplm (approved):

PR #433, opened 2021-11-24
Fix duplicte chains

maciejhirsz:
:point_up:


jsdw (approved):

dvdplm (approved):
Still not clear to me how we go from two chainx to all of them over time but let’s try this.

jsdw:
As Maciej says in the chat, passing two nodes with the same "key" basically confused react (which tries to map nodes to unique key values) and probably left it failing to clean up properly each time it had to rerender two ndoes (but only probably kept a reference to one of them)

PR #440, opened 2022-01-07
Allow chains to be sorted and filtered

### LINKED ISSUE #276: Search / filter network by name
Currently networks are sorted by node account, so it is almost impossible to locate a network with low node count. There are just no deterministic order.

Will be useful if there is a way to filter network by name.

jsdw:
And tweak various styling bits and pieces.

Here's what the new "all chains" modal looks like now:

Screenshot 2022-01-07 at 16 54 16

Screenshot 2022-01-07 at 16 54 05

It's far from perfect, but I think it'll do for now!

Closes #276

### DIFF
--- a/frontend/src/components/AllChains.tsx
+++ b/frontend/src/components/AllChains.tsx
@@ -29,46 +29,152 @@ export namespace AllChains {
   }
 }
 
-export class AllChains extends React.Component<AllChains.Props, {}> {
-  public render() {
-    const { chains, subscribed } = this.props;
-    const close = subscribed ? `#list/${subscribed}` : '#list';
+export function AllChains(props: AllChains.Props) {
+  const { chains, subscribed, connection } = props;
+  const [filterText, setFilterText] = React.useState('');
+  const [sortBy, setSortBy] = React.useState(SortBy.NumberOfNodes);
 
+  function close() {
+    window.location.hash = subscribed ? `#list/${subscribed}` : '#list';
+  }
+
+  function sortByAlphabetical() {
+    setSortBy(SortBy.Alphabetical);
+  }
+
+  function sortByNumberOfNodes() {
+    setSortBy(SortBy.NumberOfNodes);
+  }
+
+  function updateFilterText(ev: React.FormEvent<HTMLInputElement>) {
+    ev.stopPropagation();
+    setFilterText(ev.currentTarget.value);
+  }
+
+  function ignoreClicks(ev: React.MouseEvent) {
+    ev.stopPropagation();
+  }
+
+  function subscribeToChain(chain: ChainData) {
+    return () => {
+      connection.then((c) => c.subscribe(chain.genesisHash));
+      close();
+    };
+  }
+
+  const lowercaseFilterText = filterText.toLocaleLowerCase();
+  const filteredChains = chains.filter((chain) => {
+    return chain.label.toLocaleLowerCase().includes(lowercaseFilterText);
+  });
+
+  // The default sort is equal to the main display, so only sort the nodes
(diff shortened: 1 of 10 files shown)


lovelaced:
This looks great! So much easier to find things and I love the better use of space :heart:

frontend/src/components/AllChains.tsx
```diff
@@ -29,46 +29,152 @@ export namespace AllChains {
   }
 }
 
-export class AllChains extends React.Component<AllChains.Props, {}> {
-  public render() {
-    const { chains, subscribed } = this.props;
-    const close = subscribed ? `#list/${subscribed}` : '#list';
+export function AllChains(props: AllChains.Props) {
```
TarikGul, line 32: Super clean 👍

frontend/src/index.css
```diff
@@ -16,6 +16,11 @@ You should have received a copy of the GNU General Public License
 along with this program. If not, see <https://www.gnu.org/licenses/>.
 */
 
+* {
+  /* Use "standard"/common sense heights/widths (ie they include padding+boreder) */
```
TarikGul, line 20:
```suggestion
  /* Use "standard"/common sense heights/widths (ie they include padding+border) */
```

frontend/src/components/Filter.tsx
```diff
@@ -105,9 +105,14 @@ export class Filter extends React.Component<Filter.Props, {}> {
   };
 
   private onWindowKeyUp = (event: KeyboardEvent) => {
+    // Ignore if control key is being pressed
     if (event.ctrlKey) {
       return;
     }
+    // Ignore events dispatched to other elements that want to use it
+    if (['INPUT', 'TEXTAREA'].includes((event as any).target.tagName)) {
```
TarikGul, line 113: Would have thought the `React.KeyboardEvent` would have the `target` field attached to it, or atleast inherited.

jsdw: Perhaps I just couldn't find the right type (I always forget what types I need for these things); the ones I tried all assumed that target could be null, and then had very few definite props for it. I'll have another look on Monday!

jsdw: I had a bit of a look and couldn't offhand find a type that lined up with the actual properties I expect here; it looks like the react event type simply doesn't commit to much about the target (probably because it can be any element and they didn't want to guarantee any specific properties would exist in it).

I did tweak the code to be more defensive incase `target` really doesn't exist (and tagName isn't a thing)

jsdw:
I had a play with the tab animation while I was here, too :)

https://user-images.githubusercontent.com/3799193/148769020-d7d9638e-dbdc-46f6-9676-88956e1fc4c2.mov

dvdplm (approved):
Screenshots look nice, good enough for now.

Not a fan of the tab animations, perhaps something we can extract from this PR and discuss separately?

Code looks fine afaict!

frontend/src/components/Filter.tsx
TarikGul: lgtm, thanks for looking that up :) Yea I tend to forget all the types built into React often, so this was a nice little refresher.

TarikGul (approved):
lgtm 👍

jsdw:
> Screenshots look nice, good enough for now.
> 
> Not a fan of the tab animations, perhaps something we can extract from this PR and discuss separately?
> 
> Code looks fine afaict!

Happy to turn off the animation (it's a trivial change :)). My main motivation for doing this was that I wanted the tabs to feel like you could interact with them (hence the small tab raise on hover), and from there the animation to full height just made sense visually :)

PR #444, opened 2022-01-14
change network name to a mapping not string

decentration:
making PR to change the name from a string to a mapping, and correct indent.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,10 +1,8 @@
 version: "3"
 
 networks:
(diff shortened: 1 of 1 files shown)


jsdw:
This change doesn't make sense to me based on the docs at https://docs.docker.com/compose/networking/.

I've opted to just remove the network seciton entirely to ensure it doesn't cause an issue (docker will give the network a default name of its choice anyway). See #447 :)

PR #447, opened 2022-01-18
fix readme image names and remove network bit in docker-compose

### LINKED ISSUE #437: Docker image name is incomplete in example commands in readme.md, commands fail due to this
The docker image name is missing the proceeding "parity/" in step #2 and #3 under heading [Run the backend and frontend using docker-compose](https://github.com/paritytech/substrate-telemetry#run-the-backend-and-frontend-using-docker-compose)

The correct image name is: parity/substrate-telemetry-backend

If one runs the docker commands as written in step #2 and #3 in the readme.md they will fail due to the image name.  Here is the excerpt from readme.md with problem image names bolded:

Start up the backend core process. We expose port 8000 so that a UI running in a host browser can connect to the /feed endpoint.

docker run --rm -it --network=telemetry \
    --name backend-core \
    -p 8000:8000 \
    --read-only \
    **substrate-telemetry-backend** \
    telemetry_core -l 0.0.0.0:8000
In another terminal, start up the backend shard process. We tell it where it can reach the core to send messages (possible because it has been started on the same network), and we listen on and expose port 8001 so that nodes running in the host can connect and send telemetry to it.

docker run --rm -it --network=telemetry \
    --name backend-shard \
    -p 8001:8001 \
    --read-only \
    **substrate-telemetry-backend** \
    telemetry_shard -l 0.0.0.0:8001 -c http://backend-core:8000/shard_submit

jsdw:
The image names were wrong, and the network section isn't really necessary AFAICT and may have caused an issue, so I removed it.

I tested these changes by running `docker-compose up` and trying to connect a node + browser, and similarly by manually running the docker commands in the README, and both worked fine for me.

closes #437

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ If you'd like to get things running manually using Docker, you can do the follow
       --tmpfs /var/run:uid=101,gid=101 \
       --tmpfs /app/tmp:uid=101,gid=101 \
       --read-only \
-      substrate-telemetry-frontend
+      parity/substrate-telemetry-frontend
    ```
 
 With these running, you'll be able to navigate to [http://localhost:3000](http://localhost:3000) to view the UI. If you'd like to connect a node and have it send telemetry to your running shard, you can run the following:
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ If you'd like to get things running manually using Docker, you can do the follow
        --name backend-core \
        -p 8000:8000 \
        --read-only \
-       substrate-telemetry-backend \
+       parity/substrate-telemetry-backend \
        telemetry_core -l 0.0.0.0:8000
    ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -152,15 +152,14 @@ If you'd like to get things running manually using Docker, you can do the follow
    docker run --rm -it --network=telemetry \
        --name frontend \
        -p 3000:8000 \
-       --read-only \
```
niklasad1, line 155: why was this removed?

niklasad1 (approved):
LGTM

maciejhirsz (approved):

README.md
jsdw: This image writes still to `tmp`, and so the `--read-only` flag causes it to fail to start (since it's not allowed).

The "secure" version below starts a tmpfs up on `tmp` so that `--read-only` will continue to work, but it's more verbose and so it's nice to have the "simple" command here too.

PR #451, opened 2022-02-04
Off viewport tabs + favicon

maciejhirsz:
When selecting a network that's off the viewport, it will now show up on the very right with small gradient covering the regular tabs. I've changed the heuristics for the number of tabs displayed to get this working, so you might see fewer of them in normal use (the alternative is to measure when tab is out of viewport in DOM and cause layout trashing in the browser).

Also favicon, went with the plain SVG, can add the white circle underneath if we need it?

image


frontend/src/components/Chains.tsx
```diff
@@ -27,42 +28,59 @@ import './Chains.css';
...
+      this.props.subscribedHash !== nextProps.subscribedHash ||
       performance.now() - this.lastRender > RENDER_THROTTLE
     );
   }
 
   public render() {
     this.lastRender = performance.now();
+    this.subscribedChainInView = false;
 
-    const allChainsHref = this.props.subscribed
```
jsdw, line 58: Is there anything that leads to this render method being called again if the viewport width changes?

frontend/src/components/Chains.tsx
```diff
@@ -27,42 +28,59 @@ import './Chains.css';
...
-      ? `#all-chains/${this.props.subscribed}`
+    const viewportWidth = viewport().width;
+    const { chains, subscribedHash, subscribedData } = this.props;
+
+    const renderedChains = chains
+      .slice(0, (viewportWidth / AVERAGE_TAB_WIDTH) | 0)
```
jsdw, line 66: Did you mean `||` here?

maciejhirsz: No, it just forced the number to be an integer at the end.

frontend/src/components/Chains.tsx
maciejhirsz: It refreshes regularly on each render. I was thinking about adding a `resize` event listener to the component, but it seemed like an overkill.

frontend/src/components/Chains.tsx
jsdw: Ah I see!

jsdw (approved):
Love the new favicon, and showing the off screen tab is nice! 

There can be a fairly big gap between tabs on display and the "..." (esp when the selected chain isn't off the list) but I'm ok with us addressing that in a future PR if we are bothered by it (we'd need to do proper measuring I think to really nail it)

TarikGul (approved):
lgtm, great job!

PR #464, opened 2022-03-31
Add per-chain aggregate software/hardware telemetry

koute:
This is a companion for the following PR: https://github.com/paritytech/substrate/pull/11062

This adds a new tab in the telemetry GUI which will aggregate and display the software/hardware telemetry generated by `substrate`.

System information telemetry:

sysinfo

Hardware benchmark telemetry:

hwbench

Obviously for actual production chains these tables will be more richly filled out; these screenshots were taken just for my local tiny testnet with 3~4 nodes.

This was roughly inspired by the [Steam Hardware Survey](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam).

There are two types of tables here:

1) A top-N table.

   These will show the top N (currently top 10) most common values sorted according to how common they are. Those kinds of tables are currently: "Version", "Operating System", "Cpu Architecture", "CPU", "CPU Cores", "Linux Distribution" and "Linux Kernel".

   So for example, if across all of the nodes there are 100 different CPU models being used only the top 10 most popular ones will be shown, with all of the other models summed into a single "Other" row at the end of the table.

2) Ordered tables.

   These will show values ordered in a preset order, with a preset possible number of buckets. Those kinds of tables are currently: "Is Virtual Machine?", "Memory", "CPU Speed", "Memory Speed", "Disk Speed (sequential writes)" and "Disk Speed (random writes)"

   So for example, for the CPU speed benchmark there are only a preset number of rows, e.g. "Less than 0.1x", "0.1x", "0.3x", ..., "0.7x", "Baseline", "1.1x", "1.3x", ..., "2.0x", "3.0x", ..., "At least 5.0x". Each node will get assigned to one of these buckets.

If a node didn't send a given metric it will be counted in a separate "Unknown" row at the end of the table. If a table has *only* the unknown row (e.g. because none of the nodes running on a given chain support a given metric) then the table won't be shown again.

This is compatible with the current nodes running in the wild. Without my substrate PR (https://github.com/paritytech/substrate/pull/11062) there will be only these tables shown since that's what's already being sent in telemetry: "Version", "Operating System", "CPU Architecture".

*This will most likely still require some tweaking*. We might also consider adding some sort of filtering options (e.g. show only the stats for validator nodes), add the ability to download this data as a `.csv` or make it available through Grafana. Nevertheless this should be a solid MVP.

cc @emostov @ggwpez @paritytech/sdk-node


backend/telemetry_core/src/state/chain.rs
```diff
@@ -35,6 +36,298 @@ id_type! {
 pub type Label = Box<str>;
 
 const STALE_TIMEOUT: u64 = 2 * 60 * 1000; // 2 minutes
+const STATS_UPDATE_INTERVAL: Duration = Duration::from_secs(5);
```
jsdw, line 39: I haven't looked at the below code in detail, but I think that we should split it (or most of it) out into a separate file (`chain_stats.rs`?); the current file has grown quite large, and I think it's easier to read if it begins with the `Chain` struct and impl (with any other odd bits after).

jsdw, line 39: I haven't looked at the below code in detail, but I think that we should split it (or most of it) out into a separate file (`chain_stats.rs`?); the current file has grown quite large, and I think it's easier to read if it begins with the `Chain` struct and impl (with any other odd bits after).

backend/telemetry_core/src/state/chain.rs
```diff
@@ -35,6 +36,298 @@ id_type! {
...
+
+#[derive(Default)]
+struct Counter<K> {
+    map: HashMap<K, u64>,
+    empty: u64,
+}
```
jsdw, line 45: Could we have a comment for this? Is `empty` the "default" value? Perhaps this `Counter` struct should be split out into its own file, so it's super obvious what the interface it exposes is (and the internals can't be poked at)?

jsdw, line 45: Could we have a comment for this? Is `empty` the "default" value? Perhaps this `Counter` struct should be split out into its own file, so it's super obvious what the interface it exposes is (and the internals can't be poked at)?

backend/telemetry_core/src/state/chain.rs
```diff
@@ -35,6 +36,298 @@ id_type! {
...
+
+impl<K> Counter<K>
+where
+    K: Sized + std::hash::Hash + Eq,
+{
+    fn modify<'a, Q>(&mut self, key: Option<&'a Q>, increment: bool)
```
jsdw, line 51: I initially felt like having `increment` and `decrement` would be clearer, but I can sortof see why it's done this way. Instead of an `increment` boolean, how about an `enum CounterValue { Increment, Decrement }` for clarity?

jsdw, line 51: I initially felt like having `increment` and `decrement` would be clearer, but I can sortof see why it's done this way. Instead of an `increment` boolean, how about an `enum CounterValue { Increment, Decrement }` for clarity?

backend/telemetry_core/src/state/node.rs
```diff
@@ -110,6 +114,14 @@ impl Node {
         &self.best
     }
 
+    pub fn hwbench(&self) -> Option<&NodeHwBench> {
+        self.hwbench.as_ref()
+    }
+
+    pub fn replace_hwbench(&mut self, hwbench: NodeHwBench) -> Option<NodeHwBench> {
```
jsdw, line 121: just for consistency with the other update methods, how about calling this `update_hwbench`?

jsdw, line 121: just for consistency with the other update methods, how about calling this `update_hwbench`?

frontend/src/components/Stats/Stats.css
```diff
@@ -0,0 +1,74 @@
...
+.Stats {
+  text-align: center;
+  padding-top: 2.5rem;
+  padding-bottom: 0.1rem;
+}
+
+.Stats-category {
+  text-align: left;
+  background-color: #fff;
+  margin-bottom: 2.5rem;
```
jsdw, line 28: I'm not a huge fan of the thick black margin between each category, but I'm also not bothered enough to suggest an alternative! It'll be easier to see what it's like I think when it's deployed with real node data!

frontend/src/components/Stats/Stats.tsx
```diff
@@ -0,0 +1,182 @@
...
+  if (max === null) {
+    return 'More than ' + (min / 100).toFixed(1) + 'x';
+  }
+  if (min <= 100 && max >= 100) {
+    return 'Baseline';
+  }
+  return (min / 100).toFixed(1) + 'x';
+}
+
+const LIST: any[] = [
```
jsdw, line 141:
```suggestion
const LIST: [string, string, (value: any) => string][] = [
```

jsdw, line 141:
```suggestion
const LIST: [string, string, (value: any) => string][] = [
```

frontend/src/components/Stats/Stats.tsx
```diff
@@ -0,0 +1,182 @@
...
+            <th className="Stats-value">{label}</th>
+          </tr>
+        </thead>
+        <tbody>{entries}</tbody>
+      </table>
+    </div>
+  );
+}
+
+function identity(value: any): string {
```
jsdw, line 104: Is `any` the only choice, or would `string | number` work?

We could add a `type Value = string | number` and then use that in place of `any`? (I'd prefer to avoid any where we can)

jsdw, line 104: Is `any` the only choice, or would `string | number` work?

We could add a `type Value = string | number` and then use that in place of `any`? (I'd prefer to avoid any where we can)

frontend/src/components/Stats/Stats.css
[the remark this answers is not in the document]
koute: Yeah, I did it to make it easier to see where one category ends and another starts, but in general I fully expect that we might tweak this once we can see how it looks like with real data. (The number of entries in the tables will also probably need tweaking, etc.)

koute:
@jsdw Thanks for the review!

I've applied all of your suggestions; I've also improved TS typing so that there aren't any `any` anymore.

Since I've split up some of the code into their own separate files you might want to review the newest changes commit-by-commit to make it easier to review (the commits which moved the code into their own files contain only changes necessary to facilitate the move and nothing else).

jsdw (approved):
Thanks; that's great, I'm happy with this now! When it's merged to master we can deploy in a staging env to give it a better workout (may need to prod to get the rococo nodes updated).

TarikGul (approved):
Read through all the past comments and suggestions, and it all looks great. Thanks for the contribution, curious to also see how it looks with live data :)

👍

koute:
Thanks!

I took the liberty of merging it in. So I guess now it'd be a good idea to try and test it out in a staging environment? Who should I talk to to help me set that up? (:

TarikGul:
@koute So it was actually deployed to staging already and passed https://gitlab.parity.io/parity/mirrors/substrate-telemetry/-/jobs/1526662. Each time a PR is merged into master it automatically gets deployed, but I dont actually know where to access it off the top of my head. 

cc: @ArshamTeymouri Any idea what the address is?

PR #469, opened 2022-05-19
fix: eject CRA

TarikGul:
This PR focuses on ejecting CRA, and moving towards a more configurable webpack config. 

Added these two tslint rules:

```
    "member-access": false,
    "interface-over-type-literal": false,
```

Upgraded typescript to 4.7.3

This PR should be followed up with a PR to update the deps properly.


frontend/package.json
```diff
@@ -6,27 +6,61 @@
...
+    "tsconfig-paths-webpack-plugin": "^2.0.0",
+    "tslint": "^5.7.0",
+    "tslint-config-prettier": "^1.10.0",
+    "tslint-react": "^3.2.0",
+    "uglifyjs-webpack-plugin": "1.2.5",
+    "url-loader": "0.6.2",
+    "webpack": "3.8.1",
+    "webpack-dev-server": "2.11.3",
+    "webpack-manifest-plugin": "1.3.2",
+    "whatwg-fetch": "2.0.3"
```
jsdw, line 63: How much of these are actually devDependencies?

(side note; do we need the polyfills like `whatwg-fetch` nowadays? I'd have assumed not)

frontend/scripts/start.js
```diff
@@ -0,0 +1,107 @@
+'use strict';
```
jsdw, line 1: Since we already have a `start.sh`, I reckon we should rename these scripts to something like `ui_start.js`, `ui_build.js` and `ui_test.js` and then tweak said names in package.json; what do you reckon?

jsdw (approved):

jsdw:
Superceded by #521 :)

PR #477, opened 2022-06-09
Add geoip db

### LINKED ISSUE #472: Use a library for geoip functionality instead of third-party services
The codebase calls into some third-party services right now to identify location from IP address.

Those services have rate limiting and there is no way to specify API tokens for them right now, it would be much better to include a library for this purpose, especially since high precision location isn't needed anyway.

i1i1:
Closes #472 

This pr changes geo ip lookup, by using MaxMind's db. We're embedding city db in binary and it is used for a lookup.

Not sure, but it can be `lz4`-ed and it can save 41% (which is 27M).


nazar-pc (approved):
This even supports IPv6, nice!

backend/telemetry_core/src/find_location.rs
```diff
@@ -110,98 +101,25 @@ impl Locator {
...
-        let latitude = loc.next()?.ok()?;
-        let longitude = loc.next()?.ok()?;
-
-        // Guarantee that the iterator has been exhausted
-        if loc.next().is_some() {
-            return None;
-        }
-
-        Some(NodeLocation {
+        let City { city, location, .. } = self.city.lookup(ip.into()).ok()?;
```
nazar-pc, line 104: This needs to happen in a separate blocking task, it is too expensive to resolve within async context. This causes issues when deployed at scale.

backend/telemetry_core/src/find_location.rs
```diff
@@ -52,28 +50,18 @@ where
...
-                let permit = semaphore.clone().acquire_owned().await.unwrap();
                 let mut response_chan = response_chan.clone();
-                let locator = locator.clone();
+                let locator = Arc::clone(&locator);
 
-                // Once we have acquired our permit, spawn a task to avoid
-                // blocking this loop so that we can handle concurrent requests.
                 tokio::spawn(async move {
-                    let location = locator.locate(ip_address).await;
+                    let location = locator.locate(ip_address);
```
nazar-pc, line 63: More specifically, I think this is where call should be wrapped into blocking task

jsdw:
I had a look through and offhand it looks great! I'm going to try and spin up a version of telemetry in staging with this PR to see how it does; just working out how!

Thanks for the contribution!

nazar-pc:
I have a build with this PR included running at https://telemetry.subspace.network for the last day without an incidents

jsdw:
@nazar-pc ooh awesome! Do you know what is leading to some of the locations not being known in that deployment out of interest?

nazar-pc:
Not yet. It seems to be less than half right now, we had much less nodes resolved before anyway, so that is a progress. I'll take a look at it at some point later when I'm in the right mood.

My guess is those might be IP addresses used by some cloud providers that don't point to any specific location, but that is just a guess. Might also be fresh blocks of IPs allocated to some parts of the world that are not in MaxMind's database yet (database should be updated from time to time to include upcoming corrections).

i1i1:
Done, waiting for a CI approval :upside_down_face:

jsdw:
Thank you! I've had some issues getting this deployed to test internally but I am working on it :)

TarikGul (approved):
👍 From me.

jsdw (approved):
Sorry for the wait; the CI stuff didn't work and I got somewhat sidetracked. I'll merge to master; code looks good! We can test from there.

stakeworld:
@jsdw, @TarikGul, @nazar-pc , I suppose this code is also pushed to the https://telemetry.w3f.community/ backend which is used for the kusama/polkadot 1000 validator program? Dit you take in account that this push could have a lot of impact on this program since location is a part of the score?

jsdw:
That's a fair comment, thank you for raising it! I've had a chat with @wpank, and we're going to expose IP addresses directly from the telemetry feed so that things like 1kv don't have to rely directly on the locations presented. 

This stuff hasn't been released yet, so we'll just hold off on doing such a release until that information is out and Will is happy with it. 

Hopefully this'll make the location scoring better too, since location information is often quite spotty, especially via the third party APIs.

stakeworld:
@jsdw, the telemetry does seem to be updated, at least a lot of changing and missing locations, as well in the central as in the dedicated w3f telemetry? Has the update gone through? The thousand validator community is a little stressed, the changes don't seem to have reached the 1kv backend yet but if they do a lot of 1kv validators wil loose points and have to re-evaluate their locations. In our opinion it would be best to revert back to the ipinfo service or at least make sure the 1kv backend keeps using the ipinfo service. p.s. is there a matrix room to discuss these kind of things?

PR #484, opened 2022-07-21
Please consider updating this script

miloskriz:
Dockerfile current location for the frontend needs correcting


jsdw:
Those scripts seem to be leftovers from a long time ago, and aren't actually used by us any more. They also try to publish the images, so they aren't useful for others either I'd say. Let's just get rid of them entirely.

PR #486, opened 2022-07-26
Clean old scripts

miloskriz:
Following on PR #484 and following kind advise from @jsdw, this new request aims to remove the old scripts' folder and the relevant instructions of README.md


jsdw:
There's already a PR open for this; see #485 :)

Thanks for submitting though; need to get the other one reviewed, but things are a bit slow owing to holidays and such at the mo.

PR #487, opened 2022-07-29
Docker compose restart "unless stopped" policy

miloskriz:
Hello, 

This request only contains the below changes for your kind consideration:

1. Adding a restart policy to the containers so they can recover from system crashes and reboots (unless they were already stopped before the crash/reboot)
2. Correcting the indentations in two closing brackets.

Hope it helps!

cheers...

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,8 @@ services:
       - 3000:8000
     expose:
       - 3000
+    restart: unless-stopped
+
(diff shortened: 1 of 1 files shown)


jsdw (approved):
We don't really use the compose script but this change looks sensible to me; thank you!

TarikGul (approved):
LGTM, thanks for the contribution.

PR #488, opened 2022-08-02
support https and wss core url in shard's ws client

Revolution1:
I was trying to deploy telemetry to a cloud environment(azure container app), where all containers were deployed behind an API gateway. Only https endpoint were exposed. Then I found the shard's websocket client does not support tls connection. So I add one.

### DIFF
--- a/backend/common/src/ws_client/connect.rs
+++ b/backend/common/src/ws_client/connect.rs
@@ -1,3 +1,4 @@
+use std::io;
 // Source code for the Substrate Telemetry Server.
 // Copyright (C) 2021 Parity Technologies (UK) Ltd.
 //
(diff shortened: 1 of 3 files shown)


jsdw (approved):
Looks good to me, and I can see that the code follows the pattern in https://github.com/tokio-rs/tls/blob/master/tokio-rustls/examples/client/src/main.rs.

Thanks for yoru contribution!

backend/common/src/ws_client/connect.rs
```diff
@@ -1,3 +1,4 @@
+use std::io;
```
niklasad1, line 1: move this import below license header please

backend/common/src/ws_client/connect.rs
```diff
@@ -228,11 +236,18 @@ pub enum ConnectError {
...
+        port = 443
+    }
     let path = uri.path();
-
+    let port = uri.port_u16().unwrap_or(port);
     let socket = TcpStream::connect((host, port)).await?;
     socket.set_nodelay(true).expect("socket set_nodelay failed");
+    // wrap TCP stream with TLS if schema is https or wss
+    let socket =
+        may_connect_tls(socket, host.clone(), scheme == "https" || scheme == "wss").await?;
```
niklasad1, line 250: this clone seems needless, it's a &str right?

```suggestion
        may_connect_tls(socket, host, scheme == "https" || scheme == "wss").await?;
```

jsdw: Ooh good catch!

niklasad1 (approved):
LGTM, thanks

PR #489, opened 2022-08-05
Remove unused AuthoritySet bits. Fix soak test warnings

jsdw:
A couple of properties from the AuthoritySet were never actually used/passed to a node, so they have been removed to avoid confusion.

Also a wee tidy of soak tests to remove a warning.

### DIFF
--- a/backend/telemetry_core/tests/soak_tests.rs
+++ b/backend/telemetry_core/tests/soak_tests.rs
@@ -87,6 +87,7 @@ async fn run_soak_test(opts: SoakTestOpts) {
         },
         CoreOpts {
             worker_threads: opts.core_worker_threads,
+            num_aggregators: opts.core_num_aggregators,
             ..Default::default()
         },
         ShardOpts {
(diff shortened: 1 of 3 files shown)


TarikGul (approved):

backend/telemetry_core/tests/soak_tests.rs
```diff
@@ -265,7 +266,7 @@ fn chain_names(total: usize) -> impl Iterator<Item = String> {
 /// General arguments that are used to start a soak test. Run `soak_test` as
 /// instructed by its documentation for full control over what is ran, or run
 /// preconfigured variants.
-#[derive(StructOpt, Debug)]
```
niklasad1, line 268: why removed?

niklasad1 (approved):

backend/telemetry_core/tests/soak_tests.rs
jsdw: Oh I got a compiler warning that it wasn't used (presumably because the error type isn't exported and we don't debug it anywhere here). Must be a newer warning :)

PR #491, opened 2022-08-15
Expose node's IP address via feed

### LINKED ISSUE #490: Expose node IP addresses via feed
A recent contribution switched telemetry from using third party location services to an internal database.

This exposed an issue whereby the 1kv program using our location strings to help score nodes based on their location. The location DB will be more likely to return values, just because it doesn't exceed rate limits and such, butmay adversely impact that scoring.

So, let's add a CLI flag that we can start telemetry_core with to return IP addresses in the node information that comes back in the UI feed. If not provided, the address will be None.

The IP addresses will never be displayed by the feed, but this allows 1kv tooling to pick it up.

In the future we'd like a proper API for third party tooling to use to access telemetry data, but this will do the job for now, and allow the new location DB to be released while hopefully avoiding said impact.

/cc @wpank

lexnv:
The location string was previously determined via qurying an external API.
After an upstream patch, telemetry now obtains the location string via an internal
database.

The location string is utilized by the 1kv program to generate scores for nodes.

It is possible, but not likely, that ranges of IP addresses cannot be determined from
the internal database.  An empty/wrong location can adversely impact the 1kv scoring.

This patch exposes a CLI flag for the telemetry-core to expose the IP address to
the connected feeds.

While at it, fix comment typos.

### Testing Done

* The following patch is applied to obtain the submitted bytes.
```patch
backend/telemetry_core/src/aggregator/inner_loop.rs

    fn finalize_and_broadcast_to_chain_feeds(
        &mut self,
        genesis_hash: &BlockHash,
        serializer: FeedMessageSerializer,
    ) {
        if let Some(bytes) = serializer.into_finalized() {
+             info!("bytes={:?} len={:?}", bytes, bytes.len());
            self.broadcast_to_chain_feeds(genesis_hash, ToFeedWebsocket::Bytes(bytes));
        }
    }
```

* Start the telemetry core with/without the `--expose-ip` flag. 
```bash
INFO  [telemetry_core::aggregator::inner_loop] bytes=b"[3,[0,[\"ludicrous-throat-6150\",\"Parity Polkadot\",\"0.9.25-43110299e44\",null,\"12D3KooWPWB6ydrx2nPBwwWm7mtocuv6ovPEik3HzgDWGGhATsnh\",\"127.0.0.1\"],[0,0],[[]],[[],[],[]],[0,\"0x0000000000000000000000000000000000000000000000000000000000000000\",0,1660658952013,null],null,1660658945889]]" len=279

INFO  [telemetry_core::aggregator::inner_loop] bytes=b"[3,[0,[\"combative-toys-9692\",\"Parity Polkadot\",\"0.9.25-43110299e44\",null,\"12D3KooWBLmuhjsyd79zEx9DQ5FtLdyrV7DBMiP5vjA8ABBamCtX\",null],[0,0],[[]],[[],[],[]],[0,\"0x0000000000000000000000000000000000000000000000000000000000000000\",0,1660659036020,null],null,1660659027142]]" len=270
```

Closes #490.

### DIFF
--- a/backend/telemetry_core/src/aggregator/aggregator.rs
+++ b/backend/telemetry_core/src/aggregator/aggregator.rs
@@ -44,6 +44,8 @@ pub struct AggregatorOpts {
     /// How many nodes from third party chains are allowed to connect
     /// before we prevent connections from them.
     pub max_third_party_nodes: usize,
+    /// Flag to expose the node's ip address to the feed subscribers.
+    pub expose_ip: bool,
 }
 
 struct AggregatorInternal {
@@ -66,9 +68,11 @@ impl Aggregator {
 
         // Kick off a locator task to locate nodes, which hands back a channel to make location requests
         let tx_to_locator = find_location(tx_to_aggregator.clone().into_sink().with(
-            |(node_id, ip, msg)| {
+            move |(node_id, ip, msg)| {
                 future::ok::<_, flume::SendError<_>>(inner_loop::ToAggregator::FromFindLocation(
-                    node_id, ip, msg,
+                    node_id,
+                    if opts.expose_ip { Some(ip) } else { None },
+                    msg,
                 ))
             },
         ));
--- a/backend/telemetry_core/src/aggregator/aggregator.rs
+++ b/backend/telemetry_core/src/aggregator/aggregator.rs
@@ -45,7 +45,7 @@ pub struct AggregatorOpts {
     /// before we prevent connections from them.
     pub max_third_party_nodes: usize,
     /// Flag to expose the node's ip address to the feed subscribers.
-    pub expose_ip: bool,
+    pub expose_node_ips: bool,
 }
 
 struct AggregatorInternal {
@@ -67,23 +67,18 @@ impl Aggregator {
         let (tx_to_aggregator, rx_from_external) = flume::unbounded();
 
         // Kick off a locator task to locate nodes, which hands back a channel to make location requests
-        let tx_to_locator = find_location(tx_to_aggregator.clone().into_sink().with(
-            move |(node_id, ip, msg)| {
+        let tx_to_locator =
+            find_location(tx_to_aggregator.clone().into_sink().with(|(node_id, msg)| {
                 future::ok::<_, flume::SendError<_>>(inner_loop::ToAggregator::FromFindLocation(
-                    node_id,
-                    if opts.expose_ip { Some(ip) } else { None },
-                    msg,
+                    node_id, msg,
                 ))
-            },
-        ));
+            }));
 
         // Handle any incoming messages in our handler loop:
         tokio::spawn(Aggregator::handle_messages(
             rx_from_external,
             tx_to_locator,
-            opts.max_queue_len,
-            opts.denylist,
-            opts.max_third_party_nodes,
+            opts,
         ));
 
         // Return a handle to our aggregator:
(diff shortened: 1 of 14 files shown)


backend/telemetry_core/src/aggregator/aggregator.rs
```diff
@@ -44,6 +44,8 @@ pub struct AggregatorOpts {
     /// How many nodes from third party chains are allowed to connect
     /// before we prevent connections from them.
     pub max_third_party_nodes: usize,
+    /// Flag to expose the node's ip address to the feed subscribers.
+    pub expose_ip: bool,
```
jsdw, line 48: Nit: I'd call this something like "expose_node_ips" to try and make it clearer whay it does ("expose_ip" offhand could sound like it's about exposing the up of the telemetry service in some way)

jsdw (commented):
I need to have a look in more detail, but offhand I wonder why the ip address is part of the "located node" message (which comes a bit later) and not a part of whatever the initial node details message is (since I think the op would be available immediately.. We don't need to wait for any location lookups).

backend/telemetry_core/src/aggregator/aggregator.rs
lexnv: Yep, I like that name better

jsdw (approved):
Looks solid to me, nice one!

Have you tested the telemetry UI locally to make sure no errors occur etc as a result of this new IP address? (I think it's fairly likely nothing would go wrong since t he IP address is appended at the end of the array but worth a quick check if you haven't)

lexnv:
Thanks for the feedback and review!
The UI seems to not panic locally, tested with `--expose-ip` activated / disabled. 
Although the IP is appended to the end of the `NodeDetail`, it is still serialized in the middle of the message.
Not entirely sure how `ts` handles the incoming messages tho, might be worth adding a dummy IP there to ensure we don't interpret the IP bytes.
What do you think?

wpank:
This generally lgtm and serves the 1kv use case well 👍

lexnv:
Double-checked the UI's console, and no abnormal logs were reported.

backend/telemetry_core/src/aggregator/inner_loop.rs
```diff
@@ -173,24 +173,23 @@ pub struct InnerLoop {
     /// How big can the queue of messages coming in to the aggregator get before messages
     /// are prioritised and dropped to try and get back on track.
     max_queue_len: usize,
+
+    /// Flag to expose the node's ip address to the feed subscribers.
+    expose_node_ips: bool,
```
niklasad1, line 178: can this expose several IP addresses or just one?

i.e, the documentation and the field `expose_node_ips` is a bit contradicting ^^

niklasad1 (approved):

backend/telemetry_core/src/aggregator/inner_loop.rs
jsdw: As in; it exposes the IP addresses of all connected nodes :)

niklasad1: right, I still think the comment is a bit hard to understand but maybe in this context it makes sense.

lexnv: Indeed, now the comment seems a bit confusing, I guess it is coming from the plural of `ip`.

Have changed the documentation to better reflect we are referring to _all connected nodes_ as "Flag to expose the IP addresses of all connected nodes to the feed subscribers." 😄

Let me know what you think of it

niklasad1: looks great now

lexnv: Thanks for the review guys! And for making this cleaner!

PR #495, opened 2022-08-25
Tune up `backend` GHA

sergejparity:
Refactored workflow `backend.yml` by splitting it in four parts. Each job into its independent `backend_fmt.yml`, `backend_check.yml`, `backend_docs.yml` and `backend_tests.yml` respectively.

The goal of this change is to overcome limitations of the [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) GitHub Action. When multiple jobs within one workflow uses above mentioned GHA and executed in parallel it often fails to restore rust cache. Possible two solutions, make jobs using `rust-cache` to run sequentially within single workflow or setup separate workflows for every job. The latter approach is more reliable as allows to execute jobs in parallel despite little overhead  of spinning up additional runners.

As a result workflows execution time reduced almost 3 times. From 15-18 minutes to 5-7 minutes.


alvicsam (approved):

radupopa2010 (approved):

jsdw (commented):
I'm a bit sad that we need to split into multiple files to make thie work better (though well done on reducing the build times so much; that's great :)).

Could you achieve a similar thing by using one file but tagging the rust cache invocation in each job with some different thing to give each a unique cache or something?

Still; I'm ok with it either way; just wondering really because I could see this sort of change being applied across a lot of projects.

jsdw (approved):

sergejparity:
> Could you achieve a similar thing by using one file but tagging the rust cache invocation in each job with some different thing to give each a unique cache or something?

I've tried that approach as well, but looks like when it happens within one runner those caches corrupting each other. And it always a lottery whether it will work or not. So either in sequence or in separate runners.

PR #498, opened 2022-09-09
Update frontend dependencies

### LINKED ISSUE #459: Update FE dependencies
Many dependencies on the frontend (and resolutions) are very old, ~2018. We should have a proper upgrade pass and see how much of the current `package.json` can be upgraded without code changes. 
Ideally we should get rid of the `"resolutions"` section in our `package.json` file.

isSerge:
This PR includes the following changes:
- Update most npm packages to up-to-date versions
- Remove `resolutions` from `package.json`
- Replace `tslint` with `eslint` (I tried to keep the original rules)

I also removed namespace blocks from modules with class components because I experienced reference errors due to namespaces and components having same name, and removed leftover types and interfaces that are no longer used. 

I hope this should close #459


frontend/tsconfig.json
```diff
@@ -25,5 +29,5 @@
     "skipLibCheck": true
   },
   "exclude": ["node_modules", "build"],
-  "include": ["src/**/*.ts", "src/**/*.tsx", "declarations/**/*.d.ts"]
+  "include": ["src", "declarations/**/*.d.ts"]
```
TarikGul, line 32: Is there a reason for this change? Im assuming it actually just simplifies the complifier config without changing anything? As in Css files are ignored, and svgs are ignored.

TarikGul (approved):
LGTM, thanks so much for the contribution.

Side question, why were the SVG's changed?

TarikGul:
Also invloving the error, looks like the `.nvmrc` need to be updated to 14, as well as the CI node engine.

frontend/tsconfig.json
isSerge: yes, just simplification

isSerge:
@TarikGul Regarding SVGs I had this error:
`SyntaxError: unknown: Namespace tags are not supported by default. React's JSX doesn't support namespace tags. You can turn on the 'throwIfNamespace' flag to bypass this warning.`
Probably this can be adjusted somewhere in the Webpack config, but since we're using CRA, it's not exposed unless we do `eject`

niklasad1 (approved):
nice work LGTM except that the CI fails.

isSerge:
@niklasad1 it seems I need additional approval from **paritytech/ci** since this is my first contribution to this repository, how can we push this forward?

niklasad1:
@isSerge it's fine these CI changes are trivial we just need 2 approvals which we got but would be if @jsdw could have look before merging :)

sergejparity (approved):

lexnv (approved):

jsdw (approved):
Thanks for the contribution; looks great offhand!

We can test it more thoroughly when it merges so I'm up for that.

PR #504, opened 2022-10-05
Tweak logs and attempt to avoid races around removing nodes

jsdw:
- Updated some of the log messages to the nicer inline fmt style, and tweaked the odd message.
- Don't queue up "remove node" messages from shard to core if disconnected; they won't make sense after a reconnect.
- Try to avoid reusing IDs; this might cause races around disconnects if messages are queued or received with "old" IDs and new nodes are reusing said IDs.

### DIFF
--- a/backend/common/src/assign_id.rs
+++ b/backend/common/src/assign_id.rs
@@ -73,7 +73,10 @@ where
     }
 
     pub fn clear(&mut self) {
-        *self = AssignId::new();
+        // Leave the `current_id` as-is. Why? To avoid reusing IDs and risking
+        // race conditions where old messages can accidentally screw with new nodes
+        // that have been assigned the same ID.
+        self.mapping = BiMap::new();
     }
 
     pub fn iter(&self) -> impl Iterator<Item = (Id, &Details)> {
--- a/backend/common/src/assign_id.rs
+++ b/backend/common/src/assign_id.rs
@@ -47,7 +47,9 @@ where
 
     pub fn assign_id(&mut self, details: Details) -> Id {
         let this_id = self.current_id;
-        self.current_id += 1;
+        // It's very unlikely we'll ever overflow the ID limit, but in case we do,
+        // a wrapping_add will almost certainly be fine:
+        self.current_id = self.current_id.wrapping_add(1);
         self.mapping.insert(this_id, details);
         this_id.into()
     }
--- a/backend/telemetry_core/src/aggregator/inner_loop.rs
+++ b/backend/telemetry_core/src/aggregator/inner_loop.rs
@@ -248,7 +248,7 @@ impl InnerLoop {
             }
 
             if let Err(e) = metered_tx.send(msg) {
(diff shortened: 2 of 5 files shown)


lexnv (approved):

backend/telemetry_core/src/aggregator/inner_loop.rs
```diff
@@ -386,10 +386,11 @@ impl InnerLoop {
...
                     Some((node_id, _)) => node_id,
                     None => {
-                        log::error!(
-                            "Cannot find ID for node with shard/connectionId of {:?}/{:?}",
-                            shard_conn_id,
-                            local_id
+                        // It's possible that some race between removing and disconnecting shards might lead to
+                        // more than one remove message for the same node. This isn't really a problem, but we
+                        // hope it won't happen so make a note if it does:
+                        log::debug!(
```
lexnv, line 392: Are we worried about the process consuming too much memory because we may never call `remove_nodes_and_broadcast_result` here? Leading to a potentially OOM kill? 

Would it be possible for a malicious party to force the shard to submit `Remove { random_id }`?
And in that could we "mute" / disconnect the shard if it makes N wrong `Remove` calls?

jsdw: It shouldn't be possible for any party to have any control over any of the ids; we assign them all internally. 

And if the id doesn't exist for this remove message, it means the node has been removed already :)

Remove calls are also entirely internal; an external person cannot cause it to be called except by disconnecting (or when messages with certain ids go stale).

backend/common/src/assign_id.rs
```diff
@@ -73,7 +73,10 @@ where
     }
 
     pub fn clear(&mut self) {
-        *self = AssignId::new();
+        // Leave the `current_id` as-is. Why? To avoid reusing IDs and risking
+        // race conditions where old messages can accidentally screw with new nodes
+        // that have been assigned the same ID.
+        self.mapping = BiMap::new();
```
niklasad1, line 79: hmm, I don't understand.

Before we were resetting the `this.id == 0` and now just ignore this id and the next one will be `this.id +1`?
It looks like the `id` will wrap `if id + 1 > MAX` which will wrap in release mode IIRC.

Can you add an explicit panic or something so we are sure that it doesn't wrap around and ends up to replace an "old ID"?

jsdw: Yup, I did this because I worried that "old" IDs might be reused in some sorts of obscure race conditions, leading to new nodes being affected by "old" messages or something like that.

Indeed, eventually it'll overflow and panic, but since the ID is a 64 bit value it'll never realisically hit `usize::MAX` so I wasn't worried about overflowing (I almost considered adding an explicit wrapping_add but meh)

jsdw: Since it was so easy to do I did make the thing wrapping_add :)

niklasad1 (approved):

PR #514, opened 2022-11-30
fix(shard): Ignore duplicate node adds with same message ID

### LINKED ISSUE #508: The Aleph Zero telemetry page is swarmed with shadows.
When you open https://telemetry.polkadot.io/#list/0x05d5279c52c484cc80396535a316add7d47b1c5b9e0398dd1f584149341460c5 you will see lots of nodes (which are duplicates of existing ones) which are at block `0` (see the attached screenshot)
Screenshot from 2022-11-04 12-40-12

The reason why these entries showed up in the first place seems to be caused by a mistake on our side -- our node runner mistakenly attached the CLI argument `"--telemetry-url \'wss://telemetry.polkadot.io/submit/ 1\'"` twice. This is now fixed on our side, but the shadow entries don't want to disappear. This must be some backend issue, because we also run our own frontend and it looks the same as the polkadot one.

Xanewok:
Closes #508 

Verified locally by running local telemetry shard + core combo and by running
```
$ substrate --dev --telemetry-url='ws://127.0.0.1:8001/submit/ 1' --telemetry-url='ws://127.0.0.1:8001/submit/ 1'
```

Previously, two nodes would be added: one that was stuck at #0 and the other one that was being updated; killing the node did not flush the stuck one. After this patch, only a single node is added and is then removed correctly on disconnect.

cc @jsdw @DamianStraszak

### DIFF
--- a/backend/telemetry_core/tests/e2e_tests.rs
+++ b/backend/telemetry_core/tests/e2e_tests.rs
@@ -306,18 +306,11 @@ async fn e2e_feed_ignore_duplicate_nodes() {
     let (_feed_tx, mut feed_rx) = server.get_core().connect_feed().await.unwrap();
 
     let feed_messages = feed_rx.recv_feed_messages().await.unwrap();
-    assert_eq!(
-        feed_messages
-            .iter()
-            .filter(|&msg| msg
-                == &FeedMessage::AddedChain {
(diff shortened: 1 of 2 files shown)


jsdw (approved):
LGTM, lovely test!

lexnv (approved):
LGTM! 👍

PR #521, opened 2023-01-26
Update frontend build system to shed old dependencies, and address fallout from that

jsdw:
This is mainly a pass just to bring things uptodate and try to be fairly standard in the libraries we use etc.

- move to Jest for testing
- new custom (but simple) Webpack 5 builds
- 2023 copyright headers
- remove various bits that don't appear to have been used for a while
- NPM instead of yarn

I've tested this locally and all seems to work. One thing that would be good to check is that the docker builds still work OK. I think they should be fine but want to be sure that you can still pass in the SUBSTRATE_TELEMETRY_URL properly for instance.


jsdw:
Ok, so I built the docker image locally (with podman instaed of docker; ncie that this worked seamlessly) and it seems to work ok now and the SUBSTRATE_TELEMETRY_URL env var is respected.

lexnv (approved):
LGTM based on my limited ts knowledge

marshacb (approved):
lgtm!

sergejparity (approved):

Issue #535, opened 2023-05-30
Update telemetry buckets


mrcnski:
# ISSUE

## Overview

telemetry.polkadot.io is not always very helpful:

<img width="497" alt="Screenshot 2023-05-30 at 16 22 33" src="https://github.com/paritytech/substrate/assets/6035856/34007dc5-97a1-4e1c-910a-6f671bb3b6d1">

76% of validators are on kernel version "Other", which doesn't say much. :P 

## Proposal

Clean up the string (i.e. remove the `-148` part) before displaying, but allow users to view the full data somehow.

Other buckets that could be improved:

1. CPU: `60.63% | 1206 | Other`
2. Version: `22.42% | 447 | Other`
3. Memory: `39.23% | 781 | At least 64 GB` (64 GB bucket is missing and I figure that's the most common one...)

OxMik3LArTZ-Wallet:
Issue #535

mrcnski:
> Issue #535

@OxMik3LArTZ-Wallet Did you mean to link to this current issue?

OxMik3LArTZ-Wallet:
@mrcnski help wanted

mrcnski:
@OxMik3LArTZ-Wallet Please be more specific.

OxMik3LArTZ-Wallet:
@mrcnski review #550

jsdw:
Thanks for the issue @mrcnski; yeah, I agree that that's less than ideal! Possibly the string cleaning bit can be done purely in the UI. Re the CPU and memory etc, I'd have to remember what substrate provides; maybe we need to tweak something upstream, or maybe we can just add eg the 64gb bucket here :)

lexnv:
### Proposed Solution

To provide more insightful information, I've created a PR that handles the parsing in the backend entirely:

- Parsed the kernel version from `5.10.0-8-02354235`  -> `5.10.0` to avoid grouping the patch level and generic string / commit of the kernel
- Added [64; 128) GiB RAM bucket to provide more insightful info
- Added another category for the CPU vendor ( ie intel/amd/apple ) since grouping the CPUs by vendor and generation is quite specific to the CPU model name

One downside of this is that the full kernel version is no longer visible to the end user; however, the upside is that we group kernels more uniformly by their version. In other words, we'll display more kernel version numbers by loosening the grouping criteria.

### Alternatives
We could send the list of all kernel versions and CPU models to the frontend.
Then, let the frontend handle all the parsing of data.
[...]

PR #553, opened 2023-10-19
Improve reported statistics

lexnv:
This PR adds several improvements that aim to report more insightful statistical data.

### Memory Buckets

Most reported systems by our telemetry endpoint have at least 64GiB of ram. To provide more insights into the RAM of the nodes, an extra bucket is added that contains values in the range [64; 128) GiB of ram.

### Kernel Version

The linux kernel version is a string of the following format: `5.10.0-8-02354235`.

The propagated version from the backend now includes only the kernel version, major version and minor version.
The patch (` - 8 - `) and specific identified ( ` - 02354235`) are omitted for better grouping of the available versions.

### CPU Vendor

To provide more insights into the specific number of CPUs of the substrate nodes, another entry is added which groups CPUs by vendor (intel / amd).

### Package lock

Update the NPM package lock in response to updating vulnerable dependencies.

### With this PR
Deployment in staging:

Screenshot 2023-11-03 at 15 12 29
6-e6bb5330fa88)

### Next Steps

- [ ] Check if providing all information from the backend to the frontend is a viable option (ie letting the frontend do the parsing)

Closes https://github.com/paritytech/substrate-telemetry/issues/535.

cc @paritytech/subxt-team

### DIFF
--- a/backend/telemetry_core/src/state/chain_stats.rs
+++ b/backend/telemetry_core/src/state/chain_stats.rs
@@ -119,13 +119,15 @@ fn test_kernel_version_number() {
 }
 
 fn cpu_vendor(cpu: &Box<str>) -> &str {
-    if cpu.eq_ignore_ascii_case("Intel") {
+    let lowercase_cpu = cpu.to_ascii_lowercase();
+
+    if lowercase_cpu.contains("intel") {
         "Intel"
-    } else if cpu.eq_ignore_ascii_case("AMD") {
+    } else if lowercase_cpu.contains("amd") {
         "AMD"
-    } else if cpu.eq_ignore_ascii_case("ARM") {
+    } else if lowercase_cpu.contains("arm") {
         "ARM"
-    } else if cpu.eq_ignore_ascii_case("Apple") {
+    } else if lowercase_cpu.contains("apple") {
         "Apple"
     } else {
         "Other"


backend/telemetry_core/src/state/chain_stats.rs
```diff
@@ -96,6 +96,27 @@ fn bucket_memory(memory: u64) -> (u32, Option<u32>) {
...
+}
+
+#[test]
+fn test_kernel_version_number() {
+    assert_eq!(kernel_version_number(&"5.10.0-8-amd64".into()), "5.10.0");
+    assert_eq!(kernel_version_number(&"5.10.0".into()), "5.10.0");
+}
+
+fn cpu_vendor(cpu: &Box<str>) -> &str {
+    if cpu.contains("Intel") || cpu.contains("intel") {
```
niklasad1, line 114: maybe https://doc.rust-lang.org/std/string/struct.String.html#method.eq_ignore_ascii_case to keep the code a little simpler?

backend/telemetry_core/src/state/chain_stats.rs
```diff
@@ -96,6 +96,27 @@ fn bucket_memory(memory: u64) -> (u32, Option<u32>) {
         48,
         56,
         64,
+        128,
+    }
+}
+
+fn kernel_version_number(version: &Box<str>) -> &str {
+    &version[0..version.find('-').unwrap_or(version.len())]
```
niklasad1, line 104: "1.20.0+c144a98" is also a valid semver version but probably a weird edge not sure if's worth the effort to strip that one as well

linux kernel is probably not using it anyway but yeah technically possible

niklasad1: so you could just use semver::Version here :)

frontend/src/components/Stats/Stats.tsx
```diff
@@ -110,6 +110,14 @@ function formatMemory(value: Range): string {
   return min + ' GB';
 }
 
+function formatLinuxKernel(value: string): string {
+  let index = value.indexOf('-');
```
niklasad1, line 114: probably fine but see my annoying comment in the rust code as well.

backend/telemetry_core/src/state/chain_stats.rs
lexnv: By `semver::Version` you mean using another crate for this?

For now I've manually parsed the kernel version to take into account the plus sign :D

niklasad1 (approved):

jsdw (approved):
Looks good to me.

I think there is a way to deploy a specific branch to staging to test it. Let's see if that's possible before merging this to master, just to see what it's like (which might not be very informative on rococo, but just in case it shows that we should to tweak any other buckets)

PR #560, opened 2023-11-03
[DNM] testing: backend: Propagate no linux kernel versions

lexnv:


lexnv:
Testing concluded that deployments to staging work as expected, closing this.

PR #561, opened 2023-11-06
backend: Provide extensive statistics bounded to 1k

lexnv:
The purpose of this PR is to test in staging that the frontend can handle the messages exposed by the backed; as well as finding a directional limit of:
- kernel versions
- polkadot versions
- CPU names.

### DIFF
--- a/backend/telemetry_core/src/state/chain_stats.rs
+++ b/backend/telemetry_core/src/state/chain_stats.rs
@@ -189,7 +189,7 @@ impl ChainStatsCollator {
         self.linux_kernel.modify(
             sysinfo
                 .and_then(|sysinfo| sysinfo.linux_kernel.as_ref())
-                .map(kernel_version_number),
+                .map(|value| &**value),
             op,
         );
 
--- a/backend/telemetry_core/src/state/chain_stats.rs
+++ b/backend/telemetry_core/src/state/chain_stats.rs
@@ -23,6 +23,9 @@ const REFERENCE_MEMORY_SCORE: u64 = 14899;
 const REFERENCE_DISK_SEQUENTIAL_WRITE_SCORE: u64 = 485;
 const REFERENCE_DISK_RANDOM_WRITE_SCORE: u64 = 222;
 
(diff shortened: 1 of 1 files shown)


backend/telemetry_core/src/state/chain_stats.rs
```diff
@@ -246,22 +251,22 @@ impl ChainStatsCollator {
 
     pub fn generate(&self) -> ChainStats {
         ChainStats {
-            version: self.version.generate_ranking_top(10),
-            target_os: self.target_os.generate_ranking_top(10),
-            target_arch: self.target_arch.generate_ranking_top(10),
-            cpu: self.cpu.generate_ranking_top(10),
+            version: self.version.generate_ranking_top(MAXIMUM_STATS_COUNT),
```
niklasad1, line 254: Is it possible to add a few comments why 10 vs 1024 is used for certain metrics? 

FYI, I have some troubles to understand the rationale behind by looking at the code, so I would like 
to have either some kind of grouping in by two different structs in ChainStats  or a few comments.

For example:

```
struct Chainstate {
   // This contains full stats for the metrics 
   full_stats: FullStats
   // This contains the top k stats for the metrics 
   top_k_stats: TopStats
}
```



Thoughts?

lexnv: At the moment, staging does not propagate a meaningful number of statistics back to the frontend due to a limited variety of testing nodes with similar kernels and cpus.

I believe we'll end up taking a similar approach to this, but that decision still needs to be data-driven. In the meanwhile, will refactor the `ChainStats` to make it clear for which statistics we'd want more granular control in the frontend, good suggestion! Thanks!

jsdw:
@lexnv just looking through old PRs; what should we do about this one? :)

lexnv:
I'll have another look over this PR, the general idea was to provide more data to the front end.
This info was requested a while ago by the parachain team.
The next steps here would be to test it in a deployment and make sure that the UI remains response before merging this

PR #587, opened 2024-04-30
Feat: Map Node Stats (Version, Operating System etc) to each Node in Feed

CyndieKamau:
Hello, this PR is focused on providing a more detailed view of node metrics on the telemetry platform, by mapping the extra node info aggregated as a summary in `ChainStats` to specific nodes in the nodes' list table.

Some of the key changes implemented include:

## Backend

-  Modified `NodeDetails` struct to include the extra information (`target_os`, `target_arch`, `target_env`, `sysinfo`) in `common/src/node_types.rs`
- Updated the `AddedNode` payload to always expose the `sysinfo` details by default to get the additional stats (`cpu`, `memory`, `core_count`, `linux_kernel`, `linux_distro`, `is_virtual_machine`) found in `NodeSysInfo`

## Frontend

- Updated the `Node` constructor to handle the extra details within `NodeDetails`
-  Added new columns and components to display the newly available information

The objective is to provide telemetry users with deeper insights into the operational aspects of nodes. By exposing more detailed system information, users can better understand and track node operations.

PR #591, opened 2024-08-29
Feat: Map Node Stats (Version, Operating System etc) to each Node in Feed

jsdw:
This builds on #587 and addresses a couple of nits:

- Hide new oclumns by default; they take up too much space
- Default to showing '-' in new columns when no data is available.

When I tested this locally, I saw no data for a bunch of the new columns and need to understand why that is before merging.


CyndieKamau:
Hey @jsdw I can answer that for you.

So when setting up a local dev node, depending on the type of OS for your machine then the output is different.

For instance, in macOS you can only see around 3 data columns, but for Linux all the columns are filled. This is something I also noted down and informed @bLd75 about.

So this setup was configured to collect more information from a linux environment than macos, so the telemetry server will get more details from linux than macos. 

So to confirm everything was working we had to test on a linux environment.

MrishoLukamba:
> This builds on #587 and addresses a couple of nits:
> 
> * Hide new oclumns by default; they take up too much space
> * Default to showing '-' in new columns when no data is available.
> 
> When I tested this locally, I saw no data for a bunch of the new columns and need to understand why that is before merging.

We did it in a way that you have to toggle to actually display new columns. So by default they are hidden

jsdw:
> We did it in a way that you have to toggle to actually display new columns. So by default they are hidden

If you have already edited the telemetry settings, they are saved in localStorage and so it might be that that was the case for you, and so adding new things were hidden.

When I tested it, I had no settings saved and so they all showed up by default (see the commit here where I set them to false)

jsdw:
> So to confirm everything was working we had to test on a linux environment.

Ah, thankyou for this! I'll point a linux based node to it when I'm back home and have my linux laptop to hand :)

MrishoLukamba:
> > So to confirm everything was working we had to test on a linux environment.
> 
> Ah, thankyou for this! I'll point a linux based node to it when I'm back home and have my linux laptop to hand :)

any updates on this, as it is ready to be merged

jsdw:
Sorry; this fell off my radar again! We'll aim to get it reviewed and, if all is well testing a linux node pointing to it, merged, in the coming days!

frontend/src/components/List/Column/LinuxDistroColumn.tsx
```diff
@@ -0,0 +1,45 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//Column for specifying type of distro each node is running (Linux Distribution)
```
niklasad1, line 17:
```suggestion
// Column for specifying type of distro each node is running (Linux Distribution)
```

niklasad1 (approved):

frontend/src/components/List/Column/VersionColumn.tsx
```diff
@@ -0,0 +1,44 @@
...
+import * as React from 'react';
+import { ColumnProps } from './';
+import { Node } from '../../../state';
+import icon from '../../../icons/file-binary.svg';
+
+export class VersionColumn extends React.Component<ColumnProps> {
+  public static readonly label = 'version';
+  public static readonly icon = icon;
+  public static readonly width = 154;
+  public static readonly setting = 'version';
```
pkhry, line 28: Not used anywhere?

frontend/src/components/List/Column/index.ts
```diff
@@ -33,3 +33,12 @@ export * from './BlockTimeColumn';
 export * from './BlockPropagationColumn';
 export * from './LastBlockColumn';
 export * from './UptimeColumn';
+export * from './CpuArchitectureColumn'; //extra columns added
```
pkhry, line 36: Dq: but given new column follow the same structure and same pattern, can't they be factored out into one?

frontend/src/state.ts
```diff
@@ -102,7 +113,18 @@ export class Node {
...
+    const [
+      name,
+      implementation,
+      version,
+      validator,
+      networkId,
+      target_os,
+      target_arch,
+      target_env,
+      customNullField,
```
pkhry, line 125: OK, so this is the reason for undefined earlier?

frontend/src/components/List/Column/index.ts
jsdw: Yeah I'm sure we could simplify this stuff a bunch!

jsdw: Agreed

pkhry (approved):

MrishoLukamba:
wen merge? @jsdw @niklasad1 @pkhry

jsdw:
Merged :)

CyndieKamau:
Yay! Thank you @jsdw @niklasad1 @pkhry for the reviews! :)

CyndieKamau:
@jsdw I think we can safely close the first PR yes? It's still open

jsdw:
Yup; FYI I have just deployed this to live, so the new columns are now available on telemetry.polkadot.io.

Thanks for the contribution guys; nice work!

bLd75:
Great job here, very useful infos added on nodes!
Just a remark: when adding several fields, names get truncated a lot in the list. 
It seems the table was designed for a max size and name is the only adjustable column, it would be great if it was scrollable horizontally and keep full columns length.
image

jsdw:
Yup, and scrolling to the right is naff too; I added #592 for this!

PR #595, opened 2024-12-05
Fix typo incrementing instead of decrementing

clostao:
There is an error in `MostSeen::remove` method.

The issue is that for determining the label of the network it's used a voting-like system in which when nodes start sending telemetry logs also send its network information (genesis hash, network name, etc.). For each genesis hash the network name that has been sent the most times is the one to be shown.

The issue was that when a node lost connection with telemetry instead of decreasing the number of votes from the network name the node voted the service was increasing the number of votes, leading to a corruption of the voting. With the current fix, as long as the number of nodes sending the correct network name remains higher from those sending a wrong name, the correct label should be shown.

### DIFF
--- a/backend/common/src/most_seen.rs
+++ b/backend/common/src/most_seen.rs
@@ -124,7 +124,7 @@ impl<T: Hash + Eq + Clone> MostSeen<T> {
 
         // Item is in the map; not the best anyway. decrement count.
         if let Some(count) = self.others.get_mut(item) {
-            *count += 1;
+            *count = count.saturating_sub(1);
         }
         ChangeResult::NoChange
     }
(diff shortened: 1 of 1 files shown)


jsdw:
This tracks with the comment too; thanks for your fix @clostao

teor2345:
This bug got through because the current tests don’t check the value of a label that isn’t the best, which is removed multiple times, then inserted again. I’ve added a test case which would trigger this bug.

niklasad1 (approved):
nice catch, just sign the CLA bot stuff then it should be ready to go

teor2345:
> nice catch, just sign the CLA bot stuff then it should be ready to go

Done! Just took us a few days to get work approval 🙂

lexnv (approved):
Nice! Thanks for contributing! 👍

Issue #592, opened 2024-09-20
Nit: Telemetry looks weird when too much data in the list requiring you to scroll right


jsdw:
<img width="1341" alt="Screenshot 2024-09-20 at 14 46 58" src="https://github.com/user-attachments/assets/d8ce4593-93a9-407c-b4e6-2a33deeb160b">

jsdw:
I had a bit of a look and you can _mostly_ fix this by setting `overflow-x: auto` on the "List" element with the table inside it (and then perhaps preventing that from being given a fixed height). One issue here is that the horizontal scroll bar for the table needs scrolling to since the table is quite tall and scrollbars sit along the bottom. There are hacks you can use to make a scrollbar be at the top too; might be worth having this too when the table is too big (eg https://stackoverflow.com/questions/3934271/horizontal-scrollbar-on-top-and-bottom-of-table). 

Would want to test this with both not many and too many columns in the table to see that it handles things well.

dedok:
We found the same issue in our project. We propose this fix: https://github.com/paritytech/substrate-telemetry/pull/596
UI with this fix available here: https://telemetry.qfnetwork.xyz/#/0x12a5bf29b1e9f00a603993cdef83ef344ffaf68ecf096d6c4d23a42676eeaff4

dedok:
Hello, also we propose this fix: https://github.com/paritytech/substrate-telemetry/pull/600. Seems it should work much stable compared to "overlay: auto".

PR #598, opened 2025-01-15
Partial fix for overflow-y scroll issue

jsdw:
When lots of table columns are visible, scrolling to the right looks very ugly and the table cells are hard/impossible to read (see #592).

With this tweak, we match the table backgorund colour when we scroll off, and remove a box shadow which would otherwise cut down along the table.

Before:

Screenshot 2025-01-15 at 14 53 47

After:

Screenshot 2025-01-15 at 14 53 33

This is far from the perfect fix (ideally the header would fill the space etc) but improves the situation at least.

An alternative is to just set `overflow-y: auto;` on the relevant component, but the downside to this is that, when there are too many rows to fit on one page, you can no longer see the scroll bar that you need to use to scroll the table.

### DIFF
--- a/frontend/src/components/Chain/Header.css
+++ b/frontend/src/components/Chain/Header.css
@@ -24,6 +24,7 @@ along with this program. If not, see <https://www.gnu.org/licenses/>.
   color: #000;
   min-width: 1350px;
   position: relative;
+  background: linear-gradient(0deg, rgba(0,0,0,0.2) 0%, rgb(255, 255, 255) 17%), white;
 }
 
 .Header-tabs {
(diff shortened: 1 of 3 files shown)


frontend/src/components/Chain/Header.css
```diff
@@ -24,6 +24,7 @@ along with this program. If not, see <https://www.gnu.org/licenses/>.
   color: #000;
   min-width: 1350px;
   position: relative;
+  background: linear-gradient(0deg, rgba(0,0,0,0.2) 0%, rgb(255, 255, 255) 17%), white;
```
jsdw, line 27: Just to keep the gradient that the box-shadow was providing (we remove the box shadow to avoid it cutting through the table when it's too big to fit)

niklasad1 (approved):

PR #596, opened 2025-01-10
[BUG FIX] overflow does not work well

dedok:
Hello,

Our project uses substrate-telemetry with little redesign of UI. We found a bug with overflow-y (details at the end of this message), seems a table with content can't work properly if elements are off the screen.

PR contains the fix of this behavior. Actually we already use this fix (https://telemetry.qfnetwork.xyz). 
  
**How it looks now:** https://telemetry.polkadot.io/#/0x91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3: Screenshot 2025-01-10 at 17 06 47

**After applying overflow-y: auto:** https://telemetry.qfnetwork.xyz/#/0x12a5bf29b1e9f00a603993cdef83ef344ffaf68ecf096d6c4d23a42676eeaff4

Screenshot 2025-01-10 at 17 11 30
Screenshot 2025-01-10 at 17 11 34

PR #600, opened 2025-02-28
[BUG FIX] overflow does not work well #596

QuadradS:
Hello! Here I corrected the weird behavior of the table when there were many components in it. 
I also added a mobile responsive design to get rid of such bugs on small screens

### DIFF
--- a/frontend/src/App.css
+++ b/frontend/src/App.css
@@ -20,7 +20,6 @@ along with this program. If not, see <https://www.gnu.org/licenses/>.
   text-align: left;
   font-family: Roboto, Helvetica, Arial, sans-serif;
   font-size: 16px;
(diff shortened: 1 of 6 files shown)


jsdw:
Hello, thank you for the PR!

I checked it out locally but on both Firefox (top) and Chrome (bottom) I saw this (when I had one dev node pointed at it):

Screenshot 2025-03-10 at 09 41 24

When I added a second dev node the scorllbar moved down a bit, so maybe if I added enough nodes it would go to the bottom and stay where expected or something, but I wonder if we can do better!

QuadradS:
Hi, thanks for checking my work and finding the problem. I fixed it by using a minimum height for the table. Please see the result.

